PR #1, opened 2019-07-17
Initial processing of collection manifest metadata

awcrosby:
Minimal code to get json response (parsed from metadata) when calling importer with artifact.

Part of https://github.com/ansible/galaxy-dev/issues/12


awcrosby:
>     1. Please remove commented (unused) code.
Will do, this is code that will be added in over additional individual PRs but we can review those code chunks as they are added.

>     2. IIRC we discussed that importer will work with already extracted source tree. Or it'll be updated later?
In the design doc we listed both as options, I went with with importer getting the file so it could both extract and validate filename against manifest metadata. Let me know if you see a reason for changing it.

>     3. Please rename module `models` to `schema` to avoid confusion with database models.
Sure

galaxy_importer/models.py
```diff
@@ -0,0 +1,393 @@
...
+        if invalid_licenses:
+            self.value_error(
+                "Expecting 'license' to be a list of valid SPDX license "
+                "identifiers, instead found invalid license identifiers: '{}' "
+                "in 'license' value {}. "
+                "For more info, visit https://spdx.org"
+                .format(', '.join(invalid_licenses), value))
+
+    @staticmethod
+    def _is_valid_license_id(license_id, valid_license_ids):
```
cutwater: Looks like this function is generic. It can be moved to `spdx_licenses.py`.

galaxy_importer/utils/spdx_licenses.py
```diff
@@ -0,0 +1,31 @@
...
+
+log = logging.getLogger(__name__)
+
+# use get_spdx() to ref the spdx_license info, so it
+# only loaded from disk once
+_SPDX_LICENSES = None
+
+
+def load_spdx():
+    cwd = os.path.dirname(os.path.abspath(__file__))
```
cutwater: The best approach would be not to depend on physical location of this particular file and use `pkg_resources` interface for data file lookup. 

Note: From python 3.7 there is a standard package [importlib.resources](https://docs.python.org/3.7/library/importlib.html#module-importlib.resources), but we still support 3.6 I suppose.

galaxy_importer/models.py
awcrosby: Updated

galaxy_importer/utils/spdx_licenses.py
awcrosby: Updated to use `pkg_resources`

Pipfile
```diff
@@ -7,6 +7,8 @@ verify_ssl = true
 
 [packages]
 pyyaml = "*"
+semantic_version = "*"
+attrs = "*"
```
cutwater: Please use version ranges from current version to next major version instead of using star versions (e.g. `semantic_version = ">=2.6.0,<3.0"`)

galaxy_importer/utils/spdx_licenses.py
```diff
@@ -0,0 +1,49 @@
...
+
+log = logging.getLogger(__name__)
+
+# use _get_spdx() to ref the spdx_license info, so it
+# only loaded from disk once
+_SPDX_LICENSES = None
+
+
+def _load_spdx():
+    license_path = resource_filename(__name__, 'spdx_licenses.json')
```
cutwater: The point in using resource API is that python packages are not always stored as files on raw filesystem.
For this case you can use `pkg_resources.resource_stream()` function.

PR #5, opened 2019-07-25
Consolidate metadata validation to single class and add tests

awcrosby:


tests/test_collection_info.py
```diff
@@ -0,0 +1,302 @@
...
+        'readme': 'README.rst',
+        'authors': ['Bob Smith <b.smith@acme.com>'],
+        'tags': ['testcases']
+    }
+    return metadata
+
+
+def test_collection_info(collection_info):
+    res = CollectionInfo(**collection_info)
+    assert type(res) == CollectionInfo
```
ironfroggy: this test only seems to test if Python classes create Python objects, which isn't really a valuable test. Can we make some valuable assertions about the contents? If we can't, this test is probably spurious.

tests/test_collection_info.py
```diff
@@ -0,0 +1,302 @@
...
+
+
+def test_readme_req(collection_info):
+    collection_info['readme'] = ''
+    with pytest.raises(ValueError) as exc:
+        CollectionInfo(**collection_info)
+        assert "'readme' is required by galaxy" in str(exc)
```
ironfroggy: This actually needs to be outside the `with` block. When `CollectionInfo(...)` raises, it exits the block. 

Also, `exc` here is not the exception object but an `ExceptionInfo` wrapper Pytest uses. You can get the actual exception object you want to stringify with `exc.value`.

tests/test_collection_info.py
```diff
@@ -0,0 +1,302 @@
...
+    'dependent_collection',
+    [
+        'no_dot_in_collection',
+        'too.many.dots',
+        '.too.many.dots',
+        'too.many.dots.',
+        'empty_name.',
+        '.empty_namespace',
+        'a_user._leading_underscore',
+        '5tarts_with_number.gunicorn',
```
ironfroggy: Does not test for spaces, dashes, or uppercase letters, all of which are invalid but are not caught when I try to pass them.

tests/test_collection_info.py
awcrosby: updated

tests/test_collection_info.py
awcrosby: thanks, i forgot to add the `match=` pattern here

tests/test_collection_info.py
awcrosby: I added those as tests. They are caught as expected, @ironfroggy what are the values that you are using?

PR #6, opened 2019-07-31
Write cli output to file

awcrosby:


galaxy_importer/main.py
```diff
@@ -16,37 +16,62 @@
...
 
-    :param filepath: Artifact file to import.
-    """
-    json_data = collection.import_collection(filepath)
-    # TODO: make default output go to file
-    print(json_data)
+    json_data = call_importer(filepath=args.filepath)
+
+    if args.print_result:
+        pprint(json.loads(json_data))
```
cutwater: Is this for debug purposes or supposed to be used by parent program?

In any case using `pprint` for JSON formatting is a terrible idea. It is not guaranteed that `pprint` will provide a valid formatted JSON output. You should use `json.dump() / json.dumps()` with `indent=` parameter for prettifying JSON.

galaxy_importer/main.py
```diff
@@ -16,37 +16,62 @@
...
+
+    write_output_file(json_data)
 
 
 def parse_args(args):
     parser = argparse.ArgumentParser(
         description='Import collection and run linting')
     parser.add_argument(
-        '-f', '--filepath', default=None, help='Artifact to import.')
+        '-f', '--filepath',
```
cutwater: Minor \ Naming: IMO `filepath` is too long, usually `file` or `input` or `artifact` is enough.

Also, if artifact file is a required argument for importer, usually it's better to make it a positional argument.

galaxy_importer/main.py
```diff
@@ -16,37 +16,62 @@
...
+        default=None,
+        required=True,
+        help='Artifact to import.')
+    parser.add_argument(
```
cutwater: What is the behavior and use case for this parameter, when it's set to `True`?

awcrosby: For debug purposes to see full result of importer printed to the console

galaxy_importer/main.py
awcrosby: Updated

galaxy_importer/main.py
awcrosby: This is just for debugging, updated.

@cutwater do you think we should remove the `json.dumps()` from what the galaxy_importer returns and let the caller do that?: https://github.com/ansible/galaxy-importer/blob/master/galaxy_importer/collection.py#L47

cutwater: @awcrosby yes, I think it's a better option, then you can decide on formatting when writing data to the output.

galaxy_importer/main.py
cutwater: Ok, thanks. 

Minor: usually in CLI args we use dashes as word separators (e.g. `--print-result`)

PR #7, opened 2019-07-31
Add collection loader tests

awcrosby:


tests/test_collection_loader.py
```diff
@@ -0,0 +1,109 @@
...
+            'network_user.deployment': '*'
+        }
+        assert data['metadata']['repository'] is None
+        assert data['metadata']['homepage'] is None
+        assert data['metadata']['issues'] is None
+        assert data['error'] is None
+        assert data['result'] == 'completed'
+
+
+def test_manifest_fail():
```
ironfroggy: This only tests the scenario of one missing field, maybe we should use `pytest.mark.parametrize` to run this test for all the required fields and the validation messages it causes?

awcrosby: Yeah I think that's a good idea

PR #9, opened 2019-08-02
Remove filename check from importer

awcrosby:
Since pulp_ansible passes filename as a hash value instead of the name of the file uploaded (`[namespace].[name].[version].tar.gz`), this check cannot occur until edits are made to the pulp_ansible worker.


awcrosby:
@cutwater updated

PR #10, opened 2019-08-02
Edits to allow publish on pypi

awcrosby:
Added minimal fields here, can add more classifiers, etc if needed.

### DIFF
--- a/galaxy_importer/collection.py
+++ b/galaxy_importer/collection.py
@@ -24,8 +24,8 @@ import tempfile
 import attr
 import semantic_version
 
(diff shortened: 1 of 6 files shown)


galaxy_importer/collection.py
```diff
@@ -24,8 +24,8 @@
 import attr
 import semantic_version
 
-from . import exceptions as exc
-from . import schema
+from galaxy_importer import exceptions as exc
+from galaxy_importer import schema
```
cutwater: Is that really needed to publish package to PyPI?

PR #14, opened 2019-08-16
Enable galaxy_importer plugins

awcrosby:
Part of https://github.com/ansible/galaxy-dev/issues/78

### DIFF
--- a/galaxy_importer/collection.py
+++ b/galaxy_importer/collection.py
@@ -48,22 +48,22 @@ def import_collection(filepath, logger=None):
 
 
 def _import_collection(filepath, logger):
-    filename = os.path.basename(filepath)
-
     with tempfile.TemporaryDirectory() as extract_dir:
         with tarfile.open(filepath, 'r') as pkg_tar:
             pkg_tar.extractall(extract_dir)
 
-        return CollectionLoader(extract_dir, filename, logger=logger).load()
+        data = CollectionLoader(extract_dir, filepath, logger=logger).load()
+
+    return attr.asdict(data)
 
 
 class CollectionLoader(object):
     """Loads collection and content info."""
 
-    def __init__(self, path, filename, logger=None):
+    def __init__(self, path, filepath, logger=None):
         self.log = logger or default_logger
         self.path = path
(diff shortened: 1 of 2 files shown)


galaxy_importer/collection.py
```diff
@@ -47,22 +48,29 @@ def import_collection(filepath, logger=None):
...
-
     with tempfile.TemporaryDirectory() as extract_dir:
         with tarfile.open(filepath, 'r') as pkg_tar:
             pkg_tar.extractall(extract_dir)
 
-        return CollectionLoader(extract_dir, filename, logger=logger).load()
+        data = CollectionLoader(extract_dir, filepath, logger=logger).load()
+
+        _run_post_load_plugins(
+            artifact_path=filepath,
```
cutwater: Why path to archive is passed to plugin, despite it's already extracted?

awcrosby: @cutwater `ansible-test` requires the collection be in a specific working directory `{...}/ansible_collections/{namespace}/{collection}/` - passing the filepath allows the plugin to extract the artifact a second time to accommodate this

I see that `_run_post_load_plugins()` should be located out of the `extract_dir` context manager, I will update that.

cutwater: I wonder if `ansible-test` can simply create a symbolic link in a specific format to this directory?
Also this is probably a bug that should be reported to `ansible-test`, don't you think?

awcrosby: @cutwater I'll see if the symlink will work.

The working directory is a current requirement of ansible-test:
```
(vans-e) awc:collection_reqs_test$ /home/awc/code/ansible/bin/ansible-test sanity --help
ERROR: The current working directory must be at or below:

 - the Ansible source: /home/awc/code/ansible/
 - an Ansible collection: {...}/ansible_collections/{namespace}/{collection}/

Current working directory: /home/awc/code/collection_reqs_test
```

awcrosby: symlinks will not work with ansible-test since some of the underlying tests require being in the actual directory

PR #18, opened 2019-08-26
Move dependencies from pipfile to setup.py

awcrosby:


setup.cfg
```diff
@@ -14,6 +14,17 @@ classifiers =
 [options]
 packages = find:
 
+install_requires =
+    attrs>=19.1.0,<20.0
+    pyyaml>=5.1.1,<6.0
+    semantic_version>=2.6.0,<3.0
+
+[options.extras_require]
+dev =
```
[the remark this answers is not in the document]
webknjaz: @bmbouter this still needs a dependency to be published.

.gitignore
```diff
@@ -15,3 +15,5 @@ dist/
 
 # Environments
 .env
+Pipfile
```
cutwater: As Pipfile is excluded and removed from the project, no need to keep in `.gitignore`.

.travis.yml
```diff
@@ -3,7 +3,7 @@ python: 3.6
 
 install:
   - pip install pipenv
-  - pipenv install --dev
+  - pipenv install --dev '-e .[dev]'
```
cutwater: No need to use pipenv without Pipfile.

PR #20, opened 2019-08-29
Transform doc_strings options data format

awcrosby:
Changes the `options` data format inside plugin `DOCUMENTATION` strings from this:
```
"options": {
    "conf_file": {
        "description": [
            "The remote yum configuration file to use for the transaction."
        ],
        "version_added": "0.6"
    },
```

to this:
```
"options": [
    {
        "name": "conf_file",
        "description": [
            "The remote yum configuration file to use for the transaction."
        ],
        "version_added": "0.6"
    },
```

### DIFF
--- a/galaxy_importer/loaders.py
+++ b/galaxy_importer/loaders.py
@@ -16,6 +16,7 @@
 # along with Galaxy.  If not, see <http://www.apache.org/licenses/>.
 
 import abc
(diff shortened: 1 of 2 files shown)


awcrosby:
This is the only one I saw with this structure looking at a few docs examples. Others can be added if we find UI display issues.

PR #21, opened 2019-08-30
Process documentation files inside collection

awcrosby:
Find and get collection readme, files in collection `documentation` folder, and role readme. Render the markdown to html.

Closes: https://github.com/ansible/galaxy-dev/issues/37

### DIFF
--- a/galaxy_importer/collection.py
+++ b/galaxy_importer/collection.py
@@ -32,7 +32,7 @@ from galaxy_importer.utils import markup as markup_utils
 
 default_logger = logging.getLogger(__name__)
 
-DOC_FOLDER = 'documentation'
+DOCUMENTATION_DIR = 'documentation'
 
 
 def import_collection(filepath, logger=None):
@@ -132,15 +132,13 @@ class CollectionLoader(object):
 
         readme = markup_utils.get_readme_doc_file(self.path)
         if not readme:
-            msg = 'No collection readme found'
-            self.log.error(msg)
-            raise exc.ImporterError(msg)
+            raise exc.ImporterError('No collection readme found')
         rendered_readme = schema.RenderedDocFile(
             name=readme.name, html=markup_utils.get_html(readme))
 
         rendered_doc_files = []
         doc_files = markup_utils.get_doc_files(
-            os.path.join(self.path, DOC_FOLDER))
+            os.path.join(self.path, DOCUMENTATION_DIR))
         if doc_files:
             rendered_doc_files = [
                 schema.RenderedDocFile(
--- a/galaxy_importer/collection.py
+++ b/galaxy_importer/collection.py
@@ -27,11 +27,12 @@ from galaxy_importer import exceptions as exc
 from galaxy_importer.finder import ContentFinder
 from galaxy_importer import loaders
 from galaxy_importer import schema
+from galaxy_importer.utils import markup as markup_utils
 
 
 default_logger = logging.getLogger(__name__)
 
-ALLOWED_TYPES = ['text/markdown', 'text/x-rst']
+DOC_FOLDER = 'documentation'
 
 
 def import_collection(filepath, logger=None):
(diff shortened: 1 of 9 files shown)


galaxy_importer/collection.py
```diff
@@ -135,9 +130,27 @@ def _build_docs_blob(self):
             for c in self.content_objs
         ]
 
+        readme = markup_utils.get_readme_doc_file(self.path)
+        if not readme:
+            msg = 'No collection readme found'
+            self.log.error(msg)
+            raise exc.ImporterError(msg)
```
cutwater: Which part is responsible for handing this error? Maybe message can be logged there, otherwise it can easily lead to logging message in both places.

galaxy_importer/collection.py
```diff
@@ -27,11 +27,12 @@
 from galaxy_importer.finder import ContentFinder
 from galaxy_importer import loaders
 from galaxy_importer import schema
+from galaxy_importer.utils import markup as markup_utils
 
 
 default_logger = logging.getLogger(__name__)
 
-ALLOWED_TYPES = ['text/markdown', 'text/x-rst']
+DOC_FOLDER = 'documentation'
```
cutwater: Minor: I think `directory` or `dir` is more common and widely used term.

galaxy_importer/utils/markup.py
```diff
@@ -0,0 +1,123 @@
...
+
+for _ext, _type in DOCFILE_MIMETYPES.items():
+    mimetypes.add_type(_type, _ext)
+
+
+class FileSizeError(Exception):
+    pass
+
+
+def get_readme_doc_file(directory):
```
cutwater: Can you please add docstrings describing what these functions do and return as result.

tests/test_markup_utils.py
```diff
@@ -0,0 +1,171 @@
...
+        assert res is None
+
+        for doc_file in [
+            'GETTING_STARTED.md',
+            'DEEP_DIVE.md',
+            'WHOOPS.txt',
+            'EXAMPLES.md',
+        ]:
+            self.fs.create_file(os.path.join(self.directory, doc_file))
+        self.fs.create_dir(os.path.join(self.directory, 'sub_dir_to_ignore'))
```
cutwater: Please add newline to visually separate loop from code below.

tests/test_markup_utils.py
```diff
@@ -0,0 +1,171 @@
...
+
+DocFile = collections.namedtuple('DocFile', ['text', 'mimetype'])
+
+
+class FileCreateTestCase(TestCase):
+    def setUp(self):
+        self.setUpPyfakefs()
+        self.directory = '/tmp'
+
+    def tearDown(self):
```
cutwater: No need to keep empty `tearDown` method.

tests/test_markup_utils.py
```diff
@@ -0,0 +1,171 @@
...
+> NOTE: Tool 'feature' is _beta_
+
+* Item1
+* Item2
+'''
+
+DocFile = collections.namedtuple('DocFile', ['text', 'mimetype'])
+
+
+class FileCreateTestCase(TestCase):
```
cutwater: Convention is name test case class starting with `Test`.

http://doc.pytest.org/en/latest/goodpractices.html#conventions-for-python-test-discovery

tests/test_markup_utils.py
```diff
@@ -0,0 +1,171 @@
+# (c) 2012-2019, Ansible by Red Hat
```
cutwater: Please don't mix styles. You're using pytest and xUnit styles together in one test file. Please stick to one at least in file scope.

galaxy_importer/collection.py
awcrosby: I will remove the log message here, that will also match how we raise other `ImporterError` exceptions.

When these `ImporterError`s are raised, they are handled by the caller, outside of `galaxy-importer`, via change made in #17

PR #23, opened 2019-09-04
Run ansible-lint on roles

awcrosby:
Part of https://github.com/ansible/galaxy-dev/issues/13


Pipfile
```diff
@@ -16,6 +16,7 @@ semantic_version = ">=2.6.0,<3.0"
 markdown = ">=3.1.1,<4"
 bleach = ">=3.1.0,<4"
 bleach-whitelist = ">=0.0.10,<1"
+ansible-lint = ">=4.1.0,<5.0"
```
cutwater: If galaxy-importer describes dependencies in the `setup.py` file, `Pipfile` is probably not needed.
Otherwise this PR should include updated lock file as well.

galaxy_importer/loaders.py
```diff
@@ -165,9 +167,31 @@ def load(self):
...
+        proc = Popen(
+            cmd,
+            cwd=self.root,
+            encoding='utf-8',
+            stdout=PIPE,
+        )
+
+        for line in proc.stdout:
+            line_list = line.split(' ')
+            rel_path = os.path.join(*Path(line_list[0]).parts[-3:])
```
cutwater: What is the `-3` magic number?

galaxy_importer/loaders.py
```diff
@@ -165,9 +167,31 @@ def load(self):
...
+            line_list[0] = rel_path
+            line = ' '.join(line_list)
+            yield line.strip()
+
+        # returncode 1 is app exception, 0 is no linter err, 2 is linter err
+        if proc.wait() not in (0, 2):
```
cutwater: Please do not hard code exit codes if they're anything except `0` or `-1`.

Pipfile
awcrosby: Was waiting for previous pr to merge, this is now in `setup.cfg`

galaxy_importer/loaders.py
awcrosby: This makes linter message have a shortened relative path, added a code comment note for this

galaxy_importer/loaders.py
awcrosby: Added a constant here to denote the meaning of the `ansible-lint` return code

PR #26, opened 2019-09-05
Change main cli error handling

awcrosby:
In the main cli, when an error causes an import to fail, improve output:
- Only display traceback for unexpected, non ImporterError errors
- For expected ImporterErrors display for example: `The import failed, the following ImporterError was raised: ContentLoadError('No role readme found.')`

### DIFF
--- a/galaxy_importer/main.py
+++ b/galaxy_importer/main.py
@@ -21,6 +21,7 @@ import logging
 import sys
 
 from galaxy_importer import collection
+from galaxy_importer.exceptions import ImporterError
 
 
 def main(args=None):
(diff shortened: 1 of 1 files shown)


galaxy_importer/main.py
```diff
@@ -61,8 +62,12 @@ def call_importer(file):
     """
     try:
         data = collection.import_collection(file)
+    except ImporterError as e:
+        logging.info('The import failed, the following '
+                     f'ImporterError was raised: {e!r}')
```
cutwater: I think it's better to print a human readable error message or a technical traceback depending on the auditory. This message tries to be both at the same time, because of `repr()` function used and internal exception type written.

For python Exception class `repr()` function returns something like `"Exception('foo')"`, not the exception message string. To get message string you should use `str()`

awcrosby:
Closing since incorporated into #46

PR #29, opened 2019-09-11
Accept file object and filename instead of path

awcrosby:
Related to https://github.com/ansible/galaxy-api/pull/78

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,4 +8,4 @@ test:
 
 .PHONY: test/flake8
 test/flake8:
-	@flake8 galaxy_importer tests
+	@flake8 --max-line-length=99 galaxy_importer tests
--- a/galaxy_importer/collection.py
+++ b/galaxy_importer/collection.py
@@ -46,7 +46,7 @@ def import_collection(file, filename=None, logger=None):
 
 def _import_collection(file, filename, logger):
     with tempfile.TemporaryDirectory() as extract_dir:
-        with file.open() as pkg_file, tarfile.open(fileobj=pkg_file) as pkg_tar:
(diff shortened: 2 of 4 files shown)


awcrosby:
@bmbouter I decided against ducktyping and changed from file path to file obj, if pulp_ansible still needs file path let me know and I will add it in

bmbouter:
@awcrosby that sounds fine pulp_ansible can pass in the file handle [here](https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/app/tasks/collections.py#L110). Would you be able/willing to open a PR against pulp_ansible with that? You could use a `[noissue]` if you want to keep it easy since it's a small change.

Could that PR also have pulp_ansible's code use the newer version of process_collection? Right now that dep is brought in [here](https://github.com/pulp/pulp_ansible/blob/master/setup.py#L6). Either you could cut a pre-release and upgrade the version there to use it, or not cut a pre-release and have us switch to the master branch from your github directly (not from pypi). LMK if you want more details on that.

How does ^ sound so we can start using it soon?

awcrosby:
@bmbouter I can make a PR for that…

Regarding pulp_ansible using the newer version of the process_collection - that is attempted in pulp_ansible#192 and currently blocked since some `test_sync` test collections don't pass importer validation, see https://github.com/pulp/pulp_ansible/pull/192#issuecomment-530190365 … I am trying to make the tests work around this while still keeping the test executing against a collection - but if you know a quicker way to get CI passing please let me know

bmbouter:
@awcrosby oic. Yes let me see what I can do to help get those tests passing. Thank you for drawing the attention there.

PR #32, opened 2019-09-18
Don't fail import if dir inside plugin dir

awcrosby:
This PR stops failing the import if a directory is found inside a plugin directory, it now skips the subdirectory. Support for processing content inside these subdirs is noted in https://github.com/ansible/galaxy-dev/issues/79


awcrosby:
Closing as this will be dealt with when we fully support subdirs

PR #36, opened 2019-10-21
Support doc_fragments in docs_blob output

awcrosby:
- Update extract dir to be collections path via metadata names - allows `ansible-doc` to access doc_fragments
- Improve key retrieval from `ansible-doc` output
- Calculate class attrs and fully qualified content name from current path
  - Include support for content existing in subdirectories
- Call ansible-doc with updated path and fully qualified content name

Resolves: https://github.com/ansible/galaxy-dev/issues/121

### DIFF
--- a/galaxy_importer/collection.py
+++ b/galaxy_importer/collection.py
@@ -16,7 +16,6 @@
 # along with Galaxy.  If not, see <http://www.apache.org/licenses/>.
 
 from collections import namedtuple
-import json
 import logging
 import os
 from pkg_resources import iter_entry_points
@@ -50,14 +49,8 @@ def import_collection(file, filename=None, logger=None):
 
 
 def _import_collection(file, filename, logger):
-    with tarfile.open(fileobj=file, mode='r') as pkg_tar:
-        manifest_file = pkg_tar.extractfile('MANIFEST.json')
-        manifest_data = json.load(manifest_file)
-    metadata = manifest_data['collection_info']
-    sub_path = 'ansible_collections/{}/{}'.format(metadata['namespace'], metadata['name'])
-    file.seek(0)
-
     with tempfile.TemporaryDirectory() as tmp_dir:
+        sub_path = 'ansible_collections/placeholder_namespace/placeholder_name'
         extract_dir = os.path.join(tmp_dir, sub_path)
         with tarfile.open(fileobj=file, mode='r') as pkg_tar:
             pkg_tar.extractall(extract_dir)
@@ -88,6 +81,7 @@ class CollectionLoader(object):
 
     def load(self):
         self._load_collection_manifest()
+        self._rename_extract_path()
         self._check_filename_matches_manifest()
         self._check_metadata_filepaths()
         self.content_objs = list(self._load_contents())
@@ -112,6 +106,18 @@ class CollectionLoader(object):
             except ValueError as e:
                 raise exc.ManifestValidationError(str(e))
             self.metadata = data.collection_info
+
+    def _rename_extract_path(self):
+        old_ns_dir = os.path.dirname(self.path)
+        ansible_collections_dir = os.path.dirname(old_ns_dir)
+        new_ns_dir = os.path.join(ansible_collections_dir, self.metadata.namespace)
+        os.rename(old_ns_dir, new_ns_dir)
+
+        old_name_dir = os.path.join(new_ns_dir, os.path.basename(self.path))
+        new_name_dir = os.path.join(new_ns_dir, self.metadata.name)
+        os.rename(old_name_dir, new_name_dir)
+        self.path = new_name_dir
+        self.log.debug(f'Renamed extract dir to: {self.path}')
 
     def _check_filename_matches_manifest(self):
         if not self.filename:
--- a/galaxy_importer/collection.py
+++ b/galaxy_importer/collection.py
@@ -16,6 +16,7 @@
 # along with Galaxy.  If not, see <http://www.apache.org/licenses/>.
 
 from collections import namedtuple
+import json
 import logging
 import os
 from pkg_resources import iter_entry_points
@@ -49,10 +50,17 @@ def import_collection(file, filename=None, logger=None):
 
 
 def _import_collection(file, filename, logger):
-    with tempfile.TemporaryDirectory() as extract_dir:
+    with tarfile.open(fileobj=file, mode='r') as pkg_tar:
+        manifest_file = pkg_tar.extractfile('MANIFEST.json')
+        manifest_data = json.load(manifest_file)
+    metadata = manifest_data['collection_info']
+    sub_path = 'ansible_collections/{}/{}'.format(metadata['namespace'], metadata['name'])
+    file.seek(0)
+
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        extract_dir = os.path.join(tmp_dir, sub_path)
         with tarfile.open(fileobj=file, mode='r') as pkg_tar:
             pkg_tar.extractall(extract_dir)
-
         data = CollectionLoader(extract_dir, filename, logger=logger).load()
 
     _run_post_load_plugins(
--- a/galaxy_importer/loaders.py
+++ b/galaxy_importer/loaders.py
@@ -152,13 +152,15 @@ class DocStringLoader():
         json_output = self._run_ansible_doc()
 
         if not json_output:
-            return
+            return None
 
         data = json.loads(json_output)
         if not isinstance(data, dict):
-            raise exc.ImporterError('ansible-doc output not dictionary as expected')
+            self.log.error('ansible-doc output not dictionary as expected')
+            return None
         if len(data.keys()) != 1:
-            raise exc.ImporterError('ansible-doc output did not return single top-level key')
+            self.log.error('ansible-doc output did not return single top-level key')
+            return None
         data = list(data.values())[0]
         data = self._transform_doc_strings(data)
 
--- a/galaxy_importer/loaders.py
+++ b/galaxy_importer/loaders.py
@@ -63,7 +63,6 @@ class ContentLoader(metaclass=abc.ABCMeta):
         self.root = root
         self.name = self._make_name()
 
-        self.doc_strings = None
         self.description = None
         self.readme_file = None
         self.readme_html = None
@@ -117,12 +116,17 @@ class ContentLoader(metaclass=abc.ABCMeta):
 class PluginLoader(ContentLoader):
     def load(self):
         self._log_loading()
-        self.doc_strings = self._get_doc_strings()
+        doc_strings = DocStringLoader(
+            content_type=self.content_type.value,
+            path=self.tmp_dir,
+            fq_name=self.fq_name,
+            logger=self.log,
+        ).load()
 
         return schema.Content(
             name=self.path_name,
             content_type=self.content_type,
-            doc_strings=self.doc_strings,
+            doc_strings=doc_strings,
         )
 
     def _make_name(self):
@@ -132,8 +136,16 @@ class PluginLoader(ContentLoader):
         dirname_parts = Path(os.path.dirname(self.rel_path)).parts[2:]
         return '.'.join(list(dirname_parts) + [self.name])
 
-    def _get_doc_strings(self):
-        if self.content_type.value not in ANSIBLE_DOC_SUPPORTED_TYPES:
+
+class DocStringLoader():
+    def __init__(self, content_type, path, fq_name, logger=None):
+        self.content_type = content_type
+        self.path = path
+        self.fq_name = fq_name
+        self.log = logger or default_logger
+
+    def load(self):
+        if self.content_type not in ANSIBLE_DOC_SUPPORTED_TYPES:
             return None
 
         self.log.info('Getting doc strings via ansible-doc')
@@ -143,6 +155,8 @@ class PluginLoader(ContentLoader):
             return
 
         data = json.loads(json_output)
+        if not isinstance(data, dict):
+            raise exc.ImporterError('ansible-doc output not dictionary as expected')
         if len(data.keys()) != 1:
             raise exc.ImporterError('ansible-doc output did not return single top-level key')
         data = list(data.values())[0]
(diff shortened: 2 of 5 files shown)


galaxy_importer/collection.py
```diff
@@ -49,10 +50,17 @@ def import_collection(file, filename=None, logger=None):
 
 
 def _import_collection(file, filename, logger):
-    with tempfile.TemporaryDirectory() as extract_dir:
+    with tarfile.open(fileobj=file, mode='r') as pkg_tar:
+        manifest_file = pkg_tar.extractfile('MANIFEST.json')
```
chouseknecht: Check for an exception here in case `MANIFEST.json` does not exist or cannot be extracted for some reason.

galaxy_importer/collection.py
```diff
@@ -49,10 +50,17 @@ def import_collection(file, filename=None, logger=None):
...
+        manifest_data = json.load(manifest_file)
```
chouseknecht: Check for an exception here in case `MANIFEST.json` is malformed. It's possible a user may muck with the file or a bug gets introduced in the client.

galaxy_importer/collection.py
```diff
@@ -49,10 +50,17 @@ def import_collection(file, filename=None, logger=None):
...
+    metadata = manifest_data['collection_info']
+    sub_path = 'ansible_collections/{}/{}'.format(metadata['namespace'], metadata['name'])
+    file.seek(0)
```
alikins: Could this extract the archive to tmp with generic sub_path, load the metadata, then rename the sub_path dir.

Probably doesn't matter a ton. Maybe eliminates one re-open. One advantage might that the extract could just be a 'tar' subprocess which might be less picky/finicky than 'tarfile' module.

Just an idea, nothing required.

galaxy_importer/loaders.py
```diff
@@ -134,29 +170,30 @@ def handle_nested_tables(obj, table_key):
...
+            self.fq_name,
             '--type', self.content_type.value,
-            '-M', os.path.dirname(self.rel_path),
-            self.name,
-            '--json']
+            '--json',
+        ]
         self.log.debug('CMD: {}'.format(' '.join(cmd)))
-        proc = Popen(cmd, cwd=self.root, stdout=PIPE, stderr=PIPE)
+        proc = Popen(' '.join(cmd), cwd=self.root, stdout=PIPE, stderr=PIPE, shell=True)
```
alikins: Why the change to join the cmd arg list to a string here? Keeping popen cmd args in list form tends to help avoid a lot of common problems like quoting/escaping. Also helps avoid potential security issues caused by things like user provided strings doing things like fq_name = 'foo; rm -rf /important_stuff;' and creating a cmd like `env /tmp/whatever ansible-doc foo; rm -rf /important_stuff; --type module some_module_name --json`.

Especially when combined with 'shell=True'.  Why is the `shell=True` needed?

awcrosby: in getting `ansible-doc` to recognize the path environment variable and find `doc_fragments` modules, these were the changes I added. I will double check if either or both of those edits (`join` and `shell=True`) are needed.

galaxy_importer/collection.py
awcrosby: Actually that is very helpful, it also keeps the validation checks in one place.

galaxy_importer/loaders.py
```diff
@@ -72,6 +86,23 @@ def _make_name(self):
...
+        pass
+
+    def _get_fq_name(self):
+        return '{}.{}'.format(self.fq_collection_name, self.path_name)
+
+    def _get_tmp_dir(self):
+        root_parts = Path(self.root).parts
+        return os.path.join(*root_parts[:3])
+
+    def _get_fq_collection_name(self):
```
alikins: If this and some of the other _get_* methods took args instead of using instance vars (`self.root` etc), these would be easy to unit test. 



``` python
    def _get_fq_collection_name(self, root):
          root_parts = Path(root).parts
          return '{}.{}'.format(*root_parts[-2:])
```

They could also potentially be either staticmethods or module scope methods as well. 

Reducing the number of methods that use instance variables likely allows getting rid of a lot of the instances vars which can simplify maintenance (less state on each object to track, less chance of unintentionally changing them etc).

For example, it looks like `self.doc_strings`, `self.readme_html` etc are mostly just used in load(). PluginLoader shouldn't need to keep a reference to 'doc_strings', it just needs to know how to create a schema.Content() in load() and return it.

galaxy_importer/loaders.py
```diff
@@ -106,14 +141,15 @@ def _get_doc_strings(self):
 
         if not json_output:
             return
-        data = json.loads(json_output)
-        if self.name not in data.keys():
-            self.log.error(f'No "{self.name}" key in ansible-doc output')
-            return None
 
+        data = json.loads(json_output)
+        if len(data.keys()) != 1:
```
alikins: In theory, data could be a `list` without a .keys() method. Hopefully ansible-doc output won't change (again?) though. Is there a unit test where the ansible-doc output is invalid (a list, etc?). Assuming the errors/traceback bubble back up in a useful way to detect it, I wouldn't worry about type checking, but a unit test that expects an exception may be useful.

galaxy_importer/loaders.py
```diff
@@ -134,29 +170,30 @@ def handle_nested_tables(obj, table_key):
...
-            for d in data[self.name]['return']:
+            data['return'] = dict_to_named_list(ret)
+            for d in data['return']:
                 handle_nested_tables(d, table_key='contains')
 
         return data
 
     def _run_ansible_doc(self):
         cmd = [
+            'env', f'ANSIBLE_COLLECTIONS_PATHS={self.tmp_dir}',
```
alikins: Since the ansible-doc output has a 100% chance of changing randomly in the future, it may be useful to extract the ansible-doc and doc strings handling code to it's own module or class if it makes it easier to test and change.

alikins: test_loader_doc_strings.py being a separate module also seems to support that idea.

tests/test_loaders.py
```diff
@@ -67,15 +67,15 @@ def loader_module():
     return loaders.PluginLoader(
         content_type=constants.ContentType.MODULE,
         rel_path='plugins/modules/my_sample_module.py',
-        root='/tmp/tmpiskt5e2n')
+        root='/tmp/tmpiskt5e2n/ansible_collections/my_ns/my_collection')
```
alikins: The tmp path here should probably be from one of the pytest tmpfile fixtures.

A static predictable /tmp path can be dangerous (ie, something/someone else could create a  '/tmp/tmpiskt5e2n' that is link elsewhere [/etc/passwd' or '/root/.ssh' are classic examples] that would get clobbered when a dev or ci runs the unit tests).

Granted, that is not nearly as dangerous now as it used to be thanks to per user /tmp, containers, etc. But should probably be fixed.

galaxy_importer/collection.py
awcrosby: Resolved by checking `MANIFEST.json` only once which includes those validation checks

galaxy_importer/loaders.py
awcrosby: things are working without those edits, so I have removed them

tests/test_loaders.py
awcrosby: This path is only used for string manipulation, but I renamed it so it is not /tmp and more clearly a placeholder

galaxy_importer/loaders.py
awcrosby: agreed and incorporated

galaxy_importer/loaders.py
awcrosby: added a type check, since the traceback was not super helpful

galaxy_importer/loaders.py
awcrosby: agreed and refactored

galaxy_importer/loaders.py
```diff
@@ -80,44 +108,83 @@ def _validate_name(self):
...
-            key: data[self.name].get(key, None)
+            key: data.get(key, None)
             for key in ANSIBLE_DOC_KEYS
         }
 
-    def _transform_doc_strings(self, data):
+    def _run_ansible_doc(self):
+        cmd = [
+            'env', f'ANSIBLE_COLLECTIONS_PATHS={self.path}',
+            'ansible-doc',
```
alikins: Do we know the full path to where ansible-doc will be installed for the workers?  If so, specifying that here may eliminate some ambiguity.

galaxy_importer/loaders.py
```diff
@@ -80,44 +108,83 @@ def _validate_name(self):
...
         return {
-            key: data[self.name].get(key, None)
+            key: data.get(key, None)
             for key in ANSIBLE_DOC_KEYS
         }
 
-    def _transform_doc_strings(self, data):
+    def _run_ansible_doc(self):
+        cmd = [
+            'env', f'ANSIBLE_COLLECTIONS_PATHS={self.path}',
```
alikins: I'm tempted to suggest using Popen(env={}) arg here, but that requires populating the env dict with all the vars needed for ansible-doc to run. I assume it would be at least HOME, USER, LANG, LC_*, TERM. That would be a more restricted env. But that may all be redundant with the pending bwrap changes so doesn't matter a ton at the moment.

Would be handy if 'ansible-doc' had a '--collections-paths' arg, but alas.

galaxy_importer/loaders.py
```diff
@@ -80,44 +108,83 @@ def _validate_name(self):
...
         data = self._transform_doc_strings(data)
+
         return {
-            key: data[self.name].get(key, None)
+            key: data.get(key, None)
             for key in ANSIBLE_DOC_KEYS
         }
 
-    def _transform_doc_strings(self, data):
+    def _run_ansible_doc(self):
```
alikins: +1 for having the method that executes some script and returns it output separate by itself. ie, one method to run the external command and return it's output, and another to actually parse it. Tends to make it easier to test, since you can test the parsing methods with canned output without having to do too much mock gymnastics. Really helpful when the output of the ext commands start changing across versions and you need to test multiple expected outputs.

galaxy_importer/loaders.py
awcrosby: within `galaxy-dev` it should be in `/venv/bin/ansible-doc`, but `galaxy-importer` is called by `pulp-ansible` so in non-galaxy setups we won't know. also to keep supporting running from the commandline with `.main` for testing purposes we also won't know the binary path

galaxy_importer/loaders.py
awcrosby: I opted for simply setting the one env variable needed, rather than copying or curating the environment further. I wasn't thinking we needed a more restricted env here.

As for pending bwrap changes, that is planned on the calling of `ansible-test sanity`, so that would not apply here.

galaxy_importer/loaders.py
```diff
@@ -134,54 +201,45 @@ def handle_nested_tables(obj, table_key):
                 for row in obj[table_key]:
                     handle_nested_tables(row, table_key)
 
-        doc = data[self.name].get('doc')
+        doc = data.get('doc', None)
```
alikins: Suppose you could make the default return an empty dict  `doc = data.get('doc', {})`.  Would eliminate the 'if doc' condition. And `'options' in doc.keys()` could be `'options' in doc`

But if it works, I'd likely live it alone for now. If the output format of ansible-doc is changing over time, it is hard to avoid having lots of conditional checks. And if those checks are working, don't touch em (unless there are good tests for both output formats).

tests/test_collection_loader.py
```diff
@@ -68,35 +68,48 @@
 """
 
 
+@pytest.fixture
+def tmp_collection_root():
```
alikins: Could use the pytest builtin 'tmpdir' or 'tmp_path' fixtures here. (http://doc.pytest.org/en/latest/tmpdir.html)

Something vaguely like:

``` python

@pytest.fixture
def tmp_collection_root(tmpdir):
   collection_root = tmpdir.mkdir("ansible_collections").mkdir("placeholder_namespace").mkdir("placehold_name")
   return collection_root.strpath
```

But I wouldn't worry about it now. Maybe revisit the tmpdir handling and update to use the pytest fixtures later if useful.

tests/test_loaders.py
```diff
@@ -67,15 +67,31 @@ def loader_module():
...
-        root=None)
+        root='/tmp_placeholder/tmp_placeholder/ansible_collections/my_ns/my_collection')
+
+
+@pytest.fixture
+def loader_module_subdirs():
+    return loaders.PluginLoader(
+        content_type=constants.ContentType.MODULE,
+        rel_path='plugins/modules/subdir1/subdir2/my_sample_module.py',
+        root='/tmp_placeholder/tmp_placeholder/ansible_collections/my_ns/my_collection')
```
alikins: Are these paths never actually used?

awcrosby: they are used in the `.load` method where it calls class variables to string parse `self.root`, otherwise i was thinking of just setting to None

galaxy_importer/loaders.py
```diff
@@ -217,12 +275,11 @@ def _get_readme(self):
...
         if not readme:
             raise exc.ContentLoadError('No role readme found.')
-        self.readme_file = readme.name
-        self.readme_html = markup_utils.get_html(readme)
+        return readme
 
     def _get_metadata_description(self):
         self.log.info('Getting role description')
-        pass
+        return ''
```
alikins: Whats this method for? Does schema.Content() for a role require a non None description?

awcrosby: this method is for getting the role description and will be in an upcoming pr, the value will be added to the contents json... None may be better when it is empty, I will look at it more closely in that pr.

tests/test_loaders.py
alikins: So the stuff that uses it ends up getting mocked out in the tests?

galaxy_importer/loaders.py
alikins: > As for pending bwrap changes, that is planned on the calling of `ansible-test sanity`, so that would not apply here.

Ah, ok. I was thinking all of the ansible/ansible-doc/ansible-lint etc invocations were getting bwrap'ed.

tests/test_loaders.py
awcrosby: the tests either use the value defined here, or alter it if needed for the particular test

galaxy_importer/loaders.py
```diff
@@ -80,44 +108,83 @@ def _validate_name(self):
...
+            'ansible-doc',
+            self.fq_name,
+            '--type', self.content_type,
+            '--json',
+        ]
+        self.log.debug('CMD: {}'.format(' '.join(cmd)))
+        proc = Popen(cmd, cwd=self.path, stdout=PIPE, stderr=PIPE)
+        stdout, stderr = proc.communicate()
+        if proc.returncode:
+            self.log.error(f'Error running ansible-doc: {stderr}')
```
alikins: If `ansible-doc` has meaningful return codes, it might be useful to log the returncode.

Looking at ansible-code, looks like ansible-doc will likely only return 0, 1 (any AnsibleError), 5 (AnsibleParserError), or 250 (other unhandled exception). So not super useful to users, but then again, more than likely it will be you that needs to figure out why/how ansible-doc failed so might be useful ;->

PR #37, opened 2019-10-25
Get role description from role metadata, improve tests

awcrosby:
- Get role description from role metadata
- Add tests for role description, role readme, and role load

Resolves: https://github.com/ansible/galaxy-dev/issues/116

### DIFF
--- a/galaxy_importer/loaders.py
+++ b/galaxy_importer/loaders.py
@@ -23,6 +23,7 @@ import os
 from pathlib import Path
 import re
 from subprocess import Popen, PIPE
+import yaml
 
 from galaxy_importer import constants
 from galaxy_importer import exceptions as exc
@@ -37,6 +38,7 @@ ANSIBLE_DOC_SUPPORTED_TYPES = [
     'httpapi', 'inventory', 'lookup', 'shell', 'module', 'strategy', 'vars']
 ANSIBLE_DOC_KEYS = ['doc', 'metadata', 'examples', 'return']
 ANSIBLE_LINT_EXCEPTION_RETURN_CODE = 1
+ROLE_META_FILES = ['meta/main.yml', 'meta/main.yaml', 'meta.yml', 'meta.yaml']
 
 
 class ContentLoader(metaclass=abc.ABCMeta):
@@ -281,7 +283,32 @@ class RoleLoader(ContentLoader):
 
     def _get_metadata_description(self):
         self.log.info('Getting role description')
-        return ''
+        description = None
+        meta_path = self._find_metadata_file_path(self.rel_path)
+
+        if not meta_path:
+            self.log.warning('No role metadata found')
+            return description
+
+        with open(meta_path) as fp:
+            try:
+                role_metadata = yaml.safe_load(fp)
+            except Exception:
+                self.log.error('Error during parsing of role metadata')
+        try:
+            description = role_metadata['galaxy_info']['description']
+        except KeyError:
+            self.log.warning('No role description found in role metadata')
+        return description
+
+    @staticmethod
+    def _find_metadata_file_path(rel_path):
+        """Gets path to role metadata file."""
+        for file in ROLE_META_FILES:
+            meta_path = os.path.join(rel_path, file)
+            if os.path.exists(meta_path):
+                return meta_path
+        return None
 
 
 def get_loader_cls(content_type):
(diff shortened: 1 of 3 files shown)


galaxy_importer/loaders.py
```diff
@@ -281,7 +283,29 @@ def _get_readme(self):
...
+        try:
+            description = role_metadata['galaxy_info']['description']
+        except KeyError:
+            self.log.warning('No role description found in role metadata')
+        return description
+
+    @staticmethod
+    def _get_metadata_file(rel_path):
+        """Gets path to role metadata file."""
+        for file in ROLE_META_FILES:
```
alikins: Does galaxy-import need to support py2?  Under py2, 'file' is a built in object, so using it as a var name can sometimes cause odd behavior. 

For py3 only, 'file' is no longer a builtin, so should be fine.

galaxy_importer/loaders.py
```diff
@@ -281,7 +283,29 @@ def _get_readme(self):
...
-        return ''
+        description = None
+        metadata_file = self._get_metadata_file(self.rel_path)
+
+        if not metadata_file:
+            self.log.warning('No role metadata found')
+            return description
+
+        with open(metadata_file) as fp:
+            role_metadata = yaml.safe_load(fp)
```
alikins: Could maybe use some hardening if yaml.safe_load() raises exceptions.

galaxy_importer/loaders.py
```diff
@@ -281,7 +283,29 @@ def _get_readme(self):
...
+        try:
+            description = role_metadata['galaxy_info']['description']
+        except KeyError:
+            self.log.warning('No role description found in role metadata')
```
alikins: Iirc, the way the import log messages are displayed, the messages show the role that is being loaded first.
If that is true, these log messages are fine.

If it's not clear which role is being loaded, it may be useful to add some context info to the warning/error messages
(ie, 'No role description found in role metadata for role %s', self.name). But that would only be needed if that info isnt clear in the log output.

galaxy_importer/loaders.py
```diff
@@ -281,7 +283,29 @@ def _get_readme(self):
...
+        return description
+
+    @staticmethod
+    def _get_metadata_file(rel_path):
+        """Gets path to role metadata file."""
```
alikins: (this is ok as is, just comment inspired by it)
I might lean towards calling this 'find_metadata_files' or similar, since it is searching multiple candidates and may return nothing. But that's just naming stuff, and 'find_*' is just a personal idiom I sometimes use.

I might also call it 'find_metadata_file_path' or 'get_metadata_file_path' but again just a vague suggestion. I am very inconsistent in how I name vars for things like file objects, file paths, full file paths, directory names, etc. But sometimes I find it helpful to be explicit about if something is a file (object), filename, or full path to a file.

galaxy_importer/loaders.py
awcrosby: just python3

galaxy_importer/loaders.py
awcrosby: agreed, i will add this

galaxy_importer/loaders.py
awcrosby: The way log msg are displayed, user will see which role it applies to

galaxy_importer/loaders.py
awcrosby: good suggestion, explicit is better here

tests/test_loaders.py
```diff
@@ -244,3 +255,66 @@ def test_ansible_lint_exception(mocked_popen, loader_role):
...
+    res = loader_role._get_metadata_description()
+    assert res is None
+
+
+def test_get_role_readme(temp_root, loader_role):
+    loader_role.root = temp_root
+    loader_role.rel_path = ''
+
+    with pytest.raises(exc.ContentLoadError):
+        loader_role._get_readme()
```
alikins: I might split these tests into two tests, one for the null/empty case, and one for the case where it is expected to work.

ie
``` python
def test_role_loader_content_role_error_on_no_readme(temp_root, loader_role):
￼    loader_role.root = temp_root
￼    loader_role.rel_path = ''
￼
￼    with pytest.raises(exc.ContentLoadError):
￼        loader_role._get_readme()

def test_role_loader_get_readme(temp_root, loader_role):
    loader_role.root = temp_root
    loader_role.rel_path = ''
    with open(os.path.join(temp_root, 'README.md'), 'w') as fp:
￼        fp.write('This is the role readme text')
￼   readme = loader_role._get_readme()
￼   assert readme.name == 'README.md'
￼   assert readme.text == 'This is the role readme text'
```

PR #39, opened 2019-12-11
Call ansible-doc in batch by plugin type, update log format

awcrosby:
Change ansible-doc from being called one per-plugin to be called once per plugin-type, and reflect change in log message format.

Resolves https://github.com/ansible/galaxy-dev/issues/220

### DIFF
--- a/galaxy_importer/loaders.py
+++ b/galaxy_importer/loaders.py
@@ -36,6 +36,7 @@ default_logger = logging.getLogger(__name__)
 ANSIBLE_DOC_SUPPORTED_TYPES = [
     'become', 'cache', 'callback', 'cliconf', 'connection',
     'httpapi', 'inventory', 'lookup', 'shell', 'module', 'strategy', 'vars']
+ANSIBLE_DOC_PLUGIN_MAP = {'module': 'modules'}
 ANSIBLE_DOC_KEYS = ['doc', 'metadata', 'examples', 'return']
 ANSIBLE_LINT_EXCEPTION_RETURN_CODE = 1
 ROLE_META_FILES = ['meta/main.yml', 'meta/main.yaml', 'meta.yml', 'meta.yaml']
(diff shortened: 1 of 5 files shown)


alikins:
I like the idea of calling ansible-doc less often and should be a pretty noticeable performance improvement. 

I'm a little worried it may cause the classes and data structures to get kind of entangled, but that may be hard to avoid when considering the various ways gathering the doc info that are required. Probably worthwhile though if it's a nice speedup.

galaxy_importer/loaders.py
```diff
@@ -140,52 +139,66 @@ def _make_path_name(rel_path, name):
...
+            plugin_dir_name = ANSIBLE_DOC_PLUGIN_MAP.get(plugin_type, plugin_type)
+
+            plugins = self._get_plugins(os.path.join(self.path, 'plugins', plugin_dir_name))
+
+            if plugins:
+                data = self._run_ansible_doc(plugin_type, plugins)
+                data = self._process_doc_strings(data)
+                docs[plugin_type] = data
+
+        return docs
```
alikins: (This comment doesn't need to be addressed now or potentially ever, just a note while I'm looking at the code...)

If memory use is a concern, may be useful to tweak some of these for loops into iterators/generators and yield results.

I could see mem use becoming an issue at some point considering:
- It seems like folks are just going to shove tons of stuff in a few huge collections
- Some of the doc strings can get to be fairly large
- This change may mean having all the doc strings for a whole type of plugins at once (though not sure that wasn't the case before as well)
- The relatively oppressive memory constraints in openshift

galaxy_importer/loaders.py
```diff
@@ -140,52 +139,66 @@ def _make_path_name(rel_path, name):
...
+
+    def _get_plugins(self, plugin_dir):
```
alikins: From a first glance, it's a little hard to tell what this method is expected to return. Could use a comment/docstring with a quick example. Granted, it's mostly just annoying but necessary and unavoidable path name munging, and there are unit tests!

galaxy_importer/loaders.py
```diff
@@ -140,52 +139,66 @@ def _make_path_name(rel_path, name):
...
+        plugins = []
+        for root, _, files in os.walk(plugin_dir):
+            for filename in files:
+                if not filename.endswith('.py') or filename == '__init__.py':
```
alikins: Does this end up getting called on modules that could be non python?  Or would it be hitting doc_fragment paths for those cases?  Wondering because I don't see where the previous code was doing an equivalent check (though I'm guessing the new ansible support for nested plugin dirs may require this?)

galaxy_importer/loaders.py
```diff
@@ -140,52 +139,66 @@ def _make_path_name(rel_path, name):
...
+                    continue
+                file_path = os.path.join(root, filename)
+                sub_dirs = os.path.relpath(root, plugin_dir)
+
+                fq_name_parts = [self.fq_collection_name]
+                if sub_dirs and sub_dirs != '.':
+                    fq_name_parts.extend(sub_dirs.split('/'))
+                fq_name_parts.append(os.path.basename(file_path)[:-3])
```
alikins: Where's the '3' in `[:-3]` come from?  To remove '.py' from filename?

tests/test_loader_doc_strings.py
```diff
@@ -16,65 +16,60 @@
...
-            "original_message": {
-                "description": "The original name param that was passed in",
-                "type": "str"
-            }
-        }
-    }}
-"""
-
-
 @pytest.fixture
```
alikins: Does doc_string_loader() need to be a pytest.fixture?  At first glance, it seems like it could be a regular method and called from the tests.  Is there an advantage to making it a pytest.fixture?

tests/test_loader_doc_strings.py
```diff
@@ -16,65 +16,60 @@
...
+    with open(os.path.join(temp_dir, '__init__.py'), 'w'):
+        pass
+    with open(os.path.join(temp_dir, 'should_be_ignored.txt'), 'w'):
+        pass
+    with open(os.path.join(temp_dir, 'my_module.py'), 'w'):
+        pass
+    plugins = doc_string_loader._get_plugins(temp_dir)
+
+    assert plugins == ['my_namespace.my_collection.my_module']
+    shutil.rmtree(temp_dir)
```
alikins: For this case, I might cleanup the temp_dir before doing any asserts. Otherwise failed tests could leave debris behind.

May also be a place where some of the pytest temp dir fixtures could be used.

galaxy_importer/loaders.py
awcrosby: This same check occurs in the `finder` (ignoring any non-python) which builds the list of found plugins and roles. I toyed with reusing the logic between the 2 places, but it seemed to make things more complicated than this approach.

galaxy_importer/loaders.py
awcrosby: that's right

PR #42, opened 2020-01-09
Add package config and stubs for ansible test runner

awcrosby:
The PR supports https://github.com/ansible/galaxy-dev/issues/194 by adding a config to to the galaxy-importer, and to run a container with ansible-test in multiple infrastructures.


alikins:
@awcrosby Is there a galaxy-dev issue this is related to?  If so, mention it in a commit messages and pr description.

awcrosby:
@alikins I updated the pr description, ill make sure the commit that gets merged has the reference

galaxy_importer/ansible_test/runner.py
```diff
@@ -0,0 +1,58 @@
...
+        if not self.cfg.infra_pulp:
+            self._run_ansible_test_local()
+            return
+
+        image = self._build_img_with_artifact()
+
+        if self.cfg.infra_osd:
+            self._run_image_openshift_job(image)
+        else:
+            self._run_image_local(image)
```
alikins: Since this class currently does 3 or 4 things:
- decide which way run ansible test
- run ansible-test local
- build a container image
- run container image in openshift
- run container image locally

May be easier to:
- split the config based decision to it's own method (module level, or even elsewhere )
- define the AnsibleTestRunner "public" API  (ie, just the init and run())
- provides implementations for LocalAnsibleTestRunner(), OpenshiftJobTestRunner(), LocalImageTestRunner()

That may make it easier to make a plugin or entrypoint at some point.
Would also make the impl classes more focused.
Would let the impl classes be in separate modules with just the python modules they need (ie, so only ansible_test.runner.openshift_runner is the only modules that needs to import openshift modules)

galaxy_importer/config.py
```diff
@@ -0,0 +1,66 @@
...
+# along with Galaxy.  If not, see <http://www.apache.org/licenses/>.
+
+import configparser
+import os
+
+
+FILENAME = 'galaxy-importer.cfg'
+
+
+class Config(object):
```
alikins: Could the config be in yaml? May make it easier for other tools to generate/modify the config file (write support for ini files is sometimes iffy...). Also fits in the kubernetes/openshift approach. ansible (and ansible-test) itself uses a random mixmash of ini, yaml, json, etc so ini wouldn't be out of place. 

But probably doesn't really matter. Especially if the config options are small. And more so if the loading/serializing and config instance get decoupled where it would be :v:easy:v: (airquote) to change the format.

tests/test_config_runner.py
```diff
@@ -0,0 +1,114 @@
...
+        assert cfg.infra_pulp is False
+        assert cfg.infra_osd is False
+
+
+def test_runner_no_config_file(mocker):  # noqa F811
+    config.Config()._load_config()
+    ansible_test_runner = runner.AnsibleTestRunner()
+    mocker.patch.object(ansible_test_runner, '_run_ansible_test_local')
+    ansible_test_runner.run()
+    assert not ansible_test_runner._run_ansible_test_local.called
```
alikins: I would suggest moving all the tests for the config loader from galaxy_import.config and it's formats to tests/test_config.py
And moving the tests related to _import.ansible_test.runner.AnsibleTestRunner[s]
tests/ansible_test/test_runner.py.

That way the tests/test_* names match the names in the source. And the AnsibleTestRunner tests are separated and moved to a matching test file name.

awcrosby:
> I would like to see a single Config() instance created and passed around instead of trying to make it singleton-esque.

@alikins  Possible issue there... main() is only called when using standalone not when called by pulp_ansible, so `galaxy_importer/ansible_test/runner.py` would conditionally need to create the single instance and we probably want to avoid that.

Maybe this approach?: `galaxy_importer/__init__.py` does similar to your suggested approach:
```
import config
config_data = config.ConfigFile().load()
config.set_config_module_global_vars(config_data=config_data)  
```
Then other modules `import config` module and have access to module global vars.

README.md
```diff
@@ -19,3 +19,23 @@ Run parsing/validation standalone to view log output and importer result for a b
 `python -m galaxy_importer.main [collection_artifact_file]`
 
 View log output in terminal, and view the importer result in the written file `importer_result.json`
+
+
+### Configuration
+
+An optional ini configuration file is supported, the following locations are checked in this order:
```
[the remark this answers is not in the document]
awcrosby: I was following ansible.cfg, and hoping ConfigParser would be useful, but open to using yaml

galaxy_importer/config.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def _load_config(self):
+        """Set config to defaults and update with any file config."""
+
+        self._shared_state.update(self.DEFAULTS)
+
+        file_path = None
+        file_locs = [
+            f'/etc/galaxy-importer/{FILENAME}',
+            os.path.join(os.path.dirname(__file__), FILENAME),
```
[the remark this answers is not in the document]
awcrosby: developer could add for testing, also could be adding during collection explorer packaging if all bits need to be self-contained and not rely on /etc, user running in standalone mode could put a config file there but probably more likely /etc

galaxy_importer/config.py
```diff
@@ -0,0 +1,66 @@
...
+        """Set class attrs to shared state and load config if not loaded."""
+        self.__dict__ = self._shared_state
+        if not self._is_loaded:
+            self._is_loaded = True
+            self._load_config()
+
+    def _load_config(self):
+        """Set config to defaults and update with any file config."""
+
+        self._shared_state.update(self.DEFAULTS)
```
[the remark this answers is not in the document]
awcrosby: this supports multiple instances being created and ensures config loaded only once. instances are created in main() and galaxy_importer/ansible_test/runner.py

galaxy_importer/collection.py
```diff
@@ -56,6 +57,8 @@ def _import_collection(file, filename, logger):
             pkg_tar.extractall(extract_dir)
         data = CollectionLoader(extract_dir, filename, logger=logger).load()
 
+    runner.AnsibleTestRunner(logger=logger).run()
```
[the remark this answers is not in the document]
awcrosby: > Could AnsibleTestRunner be a post_load_plugin?

For entrypoint post_load_plugins, I was hoping to reduce extra packaging work (plus right now we have 1 flavor of running ansible-test in its own hub_importer_plugin repo).

> Would you ever need to run AnsibleTestRunner more than once per collection?

Right now I don't think so

galaxy_importer/ansible_test/runner.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+default_logger = logging.getLogger(__name__)
+
+
+class AnsibleTestRunner(object):
+    """Runs ansible-test on collection, dependent on config."""
+
+    def __init__(self, logger=None):
+        self.log = logger or default_logger
```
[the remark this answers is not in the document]
awcrosby: pulp_ansible passes a logger when calling collection.import_collection(), and that gets passed throughout the galaxy-importer modules

galaxy_importer/ansible_test/runner.py
awcrosby: I think splitting it up like this makes sense

tests/test_config_runner.py
awcrosby: Incorporated

README.md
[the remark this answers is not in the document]
alikins: That's more or less my assumption. In this case, I think either way conserves misery... yaml is better, but different is also harder. 

If it is working as is, that is okay with me.

galaxy_importer/config.py
```diff
@@ -0,0 +1,86 @@
...
+    }
+    IS_LOADED = False
+    CONFIG_DATA = {}
+
+    def __init__(self):
+        """Load config once and store inside class variable CONFIG_DATA.
+        Populate instance from class variable CONFIG_DATA."""
+
+        if not Config.IS_LOADED:
+            Config.IS_LOADED = True
```
alikins: Often changing class variables (IS_LOADED, CONFIG_DATA) at runtime can lead to odd and difficult to debug behavior. I'm not sure I would be able to predict the expected behavior is this class is instantiated multiple times
especially after Config.CONFIG_DATA is updated with config_file_data.  It is singleton-ish but for lack of better words, it seems kind of odd and troubleprone to me.

And again, if this is something that is required to work around some ansible-test weirdness, it may be unavoidable.

tests/test_runner.py
```diff
@@ -0,0 +1,80 @@
...
+    try:
+        dir = os.path.dirname(os.path.dirname(__file__))
+        config_file = os.path.join(dir, 'galaxy_importer', 'galaxy-importer.cfg')
+        yield config_file
+    finally:
+        os.remove(config_file)
+
+
+def test_get_runner_no_config_file():
+    config.Config.IS_LOADED = False
```
alikins: config.Config.IS_LOADED is false by default isnt it?  It shouldn't need to be set before the tests.  If it is getting changed at runtime by the config.Config.__init__() that is a symptom of the complexity of changing class variables dynamically.

galaxy_importer/collection.py
```diff
@@ -56,6 +57,10 @@ def _import_collection(file, filename, logger):
             pkg_tar.extractall(extract_dir)
         data = CollectionLoader(extract_dir, filename, logger=logger).load()
 
+    ansible_test_runner = runners.get_runner()
```
alikins: a config.Config() instance could be passed into import_collection() and passed to runners.get_runner().

That would potentially allow each collection import to use a different test runner config/context.

galaxy_importer/ansible_test/runners/__init__.py
```diff
@@ -0,0 +1,42 @@
...
+from .image import LocalImageTestRunner
+from .openshift_job import OpenshiftJobTestRunner
+
+
+default_logger = logging.getLogger(__name__)
+
+
+def get_runner():
+    """Decide which runner class to run ansible-test based on config."""
+    cfg = config.Config()
```
alikins: Could a config.Config() instance be passed into get_runner()?

tests/test_runner.py
awcrosby: the config.Config.IS_LOADED state is maintained across tests, so it is set here to false to force config to reload.

awcrosby:
I was hoping to read from config file once instead of every import to save resources and since it shouldn't change after deployment. If it's ok to read from file on each import, then we can instantiate once and pass around as you mention.

If we don't want to read from file on each import, maybe we reuse this pattern we have used across several repos:
https://github.com/ansible/galaxy-importer/blob/master/galaxy_importer/utils/spdx_licenses.py

@alikins thoughts?

alikins:
@awcrosby I think it makes sense to only read the config file once per galaxy-importer process. And probably need to only create one Config() object per galaxy-importer process. 

But creating a Config() can be decoupled from loading the config file data, so it wouldn't matter how many Config() instances exist if they are all equal. In practice, likely only need to create one Config() instance and just creating it once is easier than making it singleton-ish and makes it easier to test.

alikins:
@awcrosby I'm thinking something along the lines of https://github.com/ansible/galaxy-importer/compare/master...alikins:pr42_config_cfg  
(note: that branch is completely untested or linted etc YMMV)

PR #45, opened 2020-02-06
Add openshift job to run ansible-test in container

awcrosby:
Part of https://github.com/ansible/galaxy-dev/issues/122

Adds openshift job definition that can run depending on the `galaxy-importer` config's infrastructure parameters. This temporarily uses a static image with an artifact already included, it will be updated to use pulp-container.

### DIFF
--- a/galaxy_importer/ansible_test/runners/openshift_job.py
+++ b/galaxy_importer/ansible_test/runners/openshift_job.py
@@ -16,12 +16,144 @@
 # along with Galaxy.  If not, see <http://www.apache.org/licenses/>.
 
 
-# from galaxy_importer.ansible_test import container_build
+import os
+import requests
+import time
+from urllib3.exceptions import InsecureRequestWarning
+import uuid
+import yaml
+
+from galaxy_importer import config
+from galaxy_importer import exceptions
 from galaxy_importer.ansible_test.runners.base import BaseTestRunner
+
+
+cfg = config.Config()
+POD_CHECK_RETRIES = 120  # TODO: try to shorten once not pulling image from quay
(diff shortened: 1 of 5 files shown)


galaxy_importer/ansible_test/runners/openshift_job.py
```diff
@@ -16,12 +16,145 @@
...
+OCP_TOKEN_PATH = '/var/run/secrets/kubernetes.io/serviceaccount/token'
+
+
 class OpenshiftJobTestRunner(BaseTestRunner):
     """Run image as an openshift job."""
     def run(self):
+        # TODO: build image with pulp-container when ready
         # image = container_build.build_image_with_artifact()
-        pass
+        image = 'quay.io/awcrosby/ans-test-with-archive'
```
cutwater: Image name hardcoded.

galaxy_importer/ansible_test/runners/openshift_job.py
```diff
@@ -16,12 +16,145 @@
...
+
+        filename = os.path.join(os.path.dirname(os.path.dirname(__file__)), 'job_template.yaml')
```
cutwater: Better to use `pkg_resources` for resolving data file path within python package.

galaxy_importer/ansible_test/runners/openshift_job.py
awcrosby: updated

galaxy_importer/ansible_test/runners/openshift_job.py
awcrosby: updated

PR #46, opened 2020-02-18
Run ansible-test locally for standalone importer

awcrosby:
Resolves https://github.com/ansible/galaxy-dev/issues/261

### DIFF
--- a/galaxy_importer/ansible_test/runners/local.py
+++ b/galaxy_importer/ansible_test/runners/local.py
@@ -42,9 +42,8 @@ class LocalAnsibleTestRunner(BaseTestRunner):
             '--failure-ok',
         ]
 
-        self.log.info(
-            'Running ansible-test sanity on '
-            f'{self.metadata.namespace}-{self.metadata.name}-{self.metadata.version} ...')
+        collection_name = f'{self.metadata.namespace}-{self.metadata.name}-{self.metadata.version}'
+        self.log.info(f'Running ansible-test sanity on {collection_name} ...')
         self.log.info(f'{" ".join(cmd)}')
 
         proc = subprocess.Popen(
@@ -58,6 +57,7 @@ class LocalAnsibleTestRunner(BaseTestRunner):
         for line in proc.stdout:
             self.log.info(line.strip())
 
-        if proc.wait() != 0:
-            msg = 'An exception occurred in ansible-test sanity'
-            self.log.error(msg)
+        return_code = proc.wait()
+        if return_code != 0:
+            self.log.error('An exception occurred in {}, returncode={}, collection={}'.format(
+                ' '.join(cmd), return_code, collection_name))
--- a/galaxy_importer/ansible_test/runners/local.py
+++ b/galaxy_importer/ansible_test/runners/local.py
@@ -16,10 +16,48 @@
 # along with Galaxy.  If not, see <http://www.apache.org/licenses/>.
 
 
+import os
+import subprocess
+
 from galaxy_importer.ansible_test.runners.base import BaseTestRunner
 
 
 class LocalAnsibleTestRunner(BaseTestRunner):
     """Run ansible-test locally with --docker or using venv."""
     def run(self):
-        pass
+        version_proc = subprocess.Popen(
+            ['ansible', '--version'],
+            stdout=subprocess.PIPE,
+            encoding='utf-8',
+        )
+        self.log.info(f'Using {list(version_proc.stdout)[0].rstrip()}')
+
+        suffix = f'ansible_collections/{self.metadata.namespace}/{self.metadata.name}/'
+        collection_dir = os.path.join(self.dir, suffix)
+
+        cmd = [
+            'ansible-test', 'sanity',
+            '--docker',
+            '--color', 'yes',
+            '--failure-ok',
+        ]
+
+        self.log.info(
+            'Running ansible-test sanity on '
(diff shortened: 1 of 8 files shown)


galaxy_importer/main.py
```diff
@@ -50,6 +48,26 @@ def main(args=None):
...
 
 
+def setup_logger(cfg):
+    """Sets up logger with custom formatter."""
+    logger = logging.getLogger()
+    logger.setLevel(getattr(logging, cfg.log_level_main, 'INFO'))
+
+    ch = logging.StreamHandler(stream=sys.stdout)
+    ch.setFormatter(CustomFormatter())
+    logger.addHandler(ch)
```
alikins: [this comment doesnt require any changes just figured I'd mention it]

In theory, could do this with the default Formatter (s) by providing two handlers on the default logger. One that handles only 'INFO' level by adding a filter that drops everything except records at 'INFO'. And the other handler would handle everything else by using a filter that drops 'INFO' records. (similar to the django 'RequireDebugFalse' filter https://docs.djangoproject.com/en/3.0/topics/logging/#django.utils.log.RequireDebugFalse).

That said, I'm not sure my idea is a good idea. The main advantage is it makes it possible to further tweak the log setup with logging dictConfig. But for this case my idea would be over complicated especially since the builtin logging dictConfig isn't being used.

galaxy_importer/main.py
```diff
@@ -50,6 +48,26 @@ def main(args=None):
...
+
+
+class CustomFormatter(logging.Formatter):
+    """Formatter that does not display INFO loglevel."""
```
alikins: Not sure I understand the requirement for not showing the log levelname when it is INFO?  Is it to preserve output format with something for compatibility?

galaxy_importer/ansible_test/runners/local.py
```diff
@@ -16,10 +16,48 @@
...
+            stdout=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
+            encoding='utf-8',
+        )
+
+        for line in proc.stdout:
+            self.log.info(line.strip())
+
+        if proc.wait() != 0:
+            msg = 'An exception occurred in ansible-test sanity'
```
alikins: Use the actual cmd line that was run here instead of just 'ansible-test sanity'

ie some variant of 

``` python
self.log.error('An exception occurred in %s', ' '.join(cmd))
# or
self.log.error('An exception occurred in {}', ' '.join(cmd))
# or if py3 only
self.log.error('An exception occurred in {}', proc.args)
```

alikins: return code and/or stderr wouldn't hurt either. Though ansible-test stderr can be kind of knarly iirc, so maybe not in that case.

alikins: Oh, and possibly the collection_dir or info indicating which collection failed. Whatever info you need to debug why/what failed when someone cuts and pastes the error into an issue.

galaxy_importer/ansible_test/runners/local.py
```diff
@@ -16,10 +16,48 @@
...
+            cwd=collection_dir,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
+            encoding='utf-8',
+        )
+
+        for line in proc.stdout:
+            self.log.info(line.strip())
+
+        if proc.wait() != 0:
```
alikins: Verify the Notes at https://docs.python.org/3.7/library/subprocess.html#subprocess.Popen.wait won't be a problem. Probably not, but since 'ansible test sanity' can be long lived and verbose it could creep up. But can also just leave it as is and worry about it later if it ever comes up.

tests/test_runner.py
```diff
@@ -15,7 +15,10 @@
 # You should have received a copy of the Apache License
 # along with Galaxy.  If not, see <http://www.apache.org/licenses/>.
 
+import logging
 import os
+import subprocess
+from types import SimpleNamespace
 
 import pytest
 from pytest_mock import mocker  # noqa F401
```
alikins: Random note... the setup.cfg has a require for 'pytest_mock' but the 'official' package name is 'pytest-mock' (https://pypi.org/project/pytest-mock/) . Doesn't really matter but just noticed.

alikins: Is this import needed?  In theory pytest-mock installs a pytest plugin that provides a fixture that should get discovered when using the pytest runner.

tests/test_runner.py
```diff
@@ -83,3 +86,39 @@ def test_ansible_test_runner_run(mocker, temp_config_file):  # noqa F811
...
+    subprocess.Popen.return_value.stdout = ['stdout_result']
+    subprocess.Popen.return_value.wait.return_value = 0
+    caplog.set_level(logging.INFO)
+
+    metadata = SimpleNamespace(
+        namespace='test_ns', name='test_name', version='test_version')
+    runner = runners.local.LocalAnsibleTestRunner(metadata=metadata)
+    runner.run()
+
+    assert len(caplog.records) == 4
```
alikins: I'm not sure what all loggers caplog ends up capturing by default. But if it's root logger or similar, it may be kind of fragile to check the total number of log records since other modules could log things that don't effect normal behavior.

galaxy_importer/main.py
awcrosby: Not super important, but since we're displaying ansible-test sanity stdout, it just prevents lines that look like `INFO: ERROR: <some sanity error>`

PR #50, opened 2020-03-03
Build image in openshift for running ansible-test sanity

awcrosby:
Part of ansible/galaxy-dev#122

Uses openshift buildconfig and imagestream to build image with a collection included that is able to run `ansible-test sanity` on the collection. This also uses the download url from pulp's storage backend.

### DIFF
--- a/galaxy_importer/ansible_test/build_template.yaml
+++ b/galaxy_importer/ansible_test/build_template.yaml
@@ -14,9 +14,12 @@ spec:
       kind: ImageStreamTag
       name: {image_name}
   source:
-    dockerfile: "FROM docker-registry.default.svc:5000/buildfactory/automation-hub-ansible-test:latest\nRUN wget '{download_url}' -O /archive/archive.tar.gz"
+    dockerfile: |
+      FROM awcrosby/galaxy-ansible-test:latest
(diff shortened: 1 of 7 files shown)


galaxy_importer/ansible_test/build_template.yaml
```diff
@@ -0,0 +1,22 @@
...
+  resources:
+    limits:
+      cpu: 500m
+      memory: 512Mi
+  output:
+    to:
+      kind: ImageStreamTag
+      name: {image_name}
+  source:
+    dockerfile: "FROM docker-registry.default.svc:5000/buildfactory/automation-hub-ansible-test:latest\nRUN wget '{download_url}' -O /archive/archive.tar.gz"
```
awcrosby: @cutwater we have a few options on where the base image can be hosted... buildfactory project, automation-hub-jobs project, or outside OCP on quay, thoughts?

cutwater: 1. For automation-hub-ansible-test I think buildfactory or quay are the best options. IIRC there were some changes that enabled quay builds for buildfactory project.

2. Can you please use multiline yaml literal to shrink line width.

PR #51, opened 2020-03-03
Update ansible-test image definition

awcrosby:
Part of ansible/galaxy-dev#122

Updating the image definition used to run `ansible-test sanity` on a collection during import.

### DIFF
--- a/docker/ansible-test/Dockerfile
+++ b/docker/ansible-test/Dockerfile
@@ -23,9 +23,6 @@ RUN apt-get update -y \
     && chmod +x /entrypoint
 
 ENV HOME /
-ENV LOGNAME=user1
-ENV USER=user1
-
 USER 1000
 
 ENTRYPOINT ["/entrypoint"]
--- a/docker/ansible-test/Dockerfile
+++ b/docker/ansible-test/Dockerfile
@@ -1,4 +1,4 @@
-FROM quay.io/ansible/default-test-container
+FROM quay.io/ansible/default-test-container:1.12
 
 RUN useradd user1 \
         --uid 1000 \
--- a/docker/ansible-test/Dockerfile
+++ b/docker/ansible-test/Dockerfile
@@ -23,6 +23,9 @@ RUN apt-get update -y \
     && chmod +x /entrypoint
 
 ENV HOME /
+ENV LOGNAME=user1
+ENV USER=user1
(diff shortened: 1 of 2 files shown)


docker/ansible-test/Dockerfile
```diff
@@ -23,6 +23,9 @@ RUN apt-get update -y \
     && chmod +x /entrypoint
 
 ENV HOME /
+ENV LOGNAME=user1
+ENV USER=user1
```
cutwater: Can you please explain the purpose of these environment variable names.
Why are you setting `USER` variable explicitly

docker/ansible-test/entrypoint.sh
```diff
@@ -5,22 +5,25 @@ set -e
...
-cp /archive/*.tar.gz .
-tar -xzf *.tar.gz
+echo "Copying and extracting collection archive..."
+cp /archive/archive.tar.gz .
+tar -xzf archive.tar.gz
 
 # Rename placeholders in path
 NAMESPACE=$(python3 -c "import json; f = open('MANIFEST.json'); namespace = json.load(f)['collection_info']['namespace']; print(namespace); f.close")
 NAME=$(python3 -c "import json; f = open('MANIFEST.json'); name = json.load(f)['collection_info']['name']; print(name); f.close")
+VERSION=$(python3 -c "import json; f = open('MANIFEST.json'); version = json.load(f)['collection_info']['version']; print(version); f.close")
```
cutwater: You are not calling `f.close()`.

cutwater:
```
read NAMESPACE NAME VERSION < <(python3 <<EOF
import json
with open('MANIFEST.json') as fp:
    ci = json.load(fp)['collection_info']
values = ci['namespace'], ci['name'], ci['version']
print(*values)
EOF
)
```

docker/ansible-test/Dockerfile
awcrosby: I was trying to resolve an issue where a python lib (ansible-doc called by ansible-test sanity) is calling `getpass.getuser()` throwing a `getpwuid(): uid not found` error since the OCP container is rootless and assigned a random user id that doesn't match any users in the pod's os. But this did not resolve it, I will keep working this issue separately.

PR #60, opened 2020-04-29
Fix openshift container import fails - 'terminated' keyerror

awcrosby:
Only look for `terminated` reason if the container does indeed have a status of `Failed`

Works https://github.com/ansible/galaxy_ng/issues/130

### DIFF
--- a/CHANGES/130.bugfix
+++ b/CHANGES/130.bugfix
@@ -1,0 +1,1 @@
+Fix openshift container import fails on checking sanity container status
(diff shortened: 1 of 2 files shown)


chouseknecht:
LGTM

CHANGES/.gitignore
```diff
@@ -0,0 +1,2 @@
+!.gitignorie
```
cutwater, line 1: I guess this is to add empty directory to `git`?
I've seen another pattern, adding `.gitkeep` file to git.

awcrosby: yes it's to retain the directory when empty. yeah `.gitkeep` seems better

PR #62, opened 2020-05-06
Build and run ansible-test via pulp-container

bmclaughlin:
Enhancement to build and run ansible-test via pulp-container for process isolation.
Will be refactoring to use `pulp-container-client` in a follow-up PR.
Tests are in-progress (current reason for Travis failing).

To test:
1. Install `galaxy-importer` into your venv.
`pip install -e .[dev]`
2. Update configuration file to include:
`RUN_ANSIBLE_TEST = True`
`INFRA_PULP = True`
3. Run `galaxy-importer` stand-alone mode:
`python -m galaxy_importer.main [collection_artifact_file]`

Issues:
  - https://github.com/ansible/galaxy_ng/issues/47
  - https://github.com/ansible/galaxy_ng/issues/81


galaxy_importer/ansible_test/builders/pulp_build.py
```diff
@@ -15,7 +15,271 @@
...
+class Build(object):
     """Use pulp-container to build ansible-test image with artifact inside."""
-    return ''
+    def __init__(self, pulp_artifact_file, api_url, content_url, metadata, logger):
+        self.api_url = api_url
+        self.content_url = content_url
+        self.basic_auth = ('admin', 'admin')
+        self.pulp_artifact_file = pulp_artifact_file
+        self.metadata = metadata
+        self.shared_dockerfile = os.path.join(
```
awcrosby, line 46: can we point to the image on quay rather than the Dockerfile defined outside `galaxy_importer`? This will not be available when you pip install `galaxy-importer` - it's only available on cloning the repo

galaxy_importer/ansible_test/builders/pulp_build.py
```diff
@@ -15,7 +15,271 @@
...
+    def _get_collection_file(self):
+        path = os.path.join(
+            os.path.dirname(os.path.realpath(__file__)),
+            'archive.tar.bz')
+        try:
+            url = Build._get_pulp_archive_url(self.pulp_artifact_file)
+            with requests.get(url, stream=True) as r:
+                with open(path, 'wb') as f:
+                    copyfileobj(r.raw, f)
+        except AttributeError:
```
awcrosby, line 67: there is a lot of logic in this exception, is this an expected code path? when would the exception occur / not occur?

galaxy_importer/ansible_test/builders/pulp_build.py
```diff
@@ -15,7 +15,271 @@
...
+        self.collection_file = self._get_collection_file()
+        self.registry_base_path = 'galaxy-importer'
+        self.registry_name = 'archive-test'
+        self.log = logger or default_logger
+        self.setup_dockerfile()
+
+    def _get_collection_file(self):
+        path = os.path.join(
+            os.path.dirname(os.path.realpath(__file__)),
+            'archive.tar.bz')
```
awcrosby, line 61: other places state `archive.tar.gz`, is the `bz` intentional?

galaxy_importer/ansible_test/builders/pulp_build.py
```diff
@@ -15,7 +15,271 @@
...
+                raise exceptions.AnsibleTestError(
+                    f'Could not create image: \
+                        {r.status_code} {r.reason} {r.text}')
+            elif status == 'completed':
+                self.log.info('Image successfully built')
+                return
+            time.sleep(API_CHECK_DELAY_SECONDS)
+        if i >= API_CHECK_DELAY_SECONDS:
+            self.cleanup()
+            raise exceptions.AnsibleTestError('Timed out building image')
```
awcrosby, line 134:
```suggestion
        self.cleanup()
        raise exceptions.AnsibleTestError('Timed out building image')
```

awcrosby: i don't think the comparison of `API_CHECK_RETRIES` to `API_CHECK_DELAY_SECONDS` yields any useful information and is always true when `API_CHECK_DELAY_SECONDS` equals 1

galaxy_importer/ansible_test/builders/pulp_build.py
```diff
@@ -15,7 +15,271 @@
...
+            path = f'home/{os.getlogin()}/.config/containers'.split('/')
+            for index, item in enumerate(path):
+                current_path = Build._get_current_path(index, path)
+                if not os.path.exists(current_path):
+                    subprocess.run(['mkdir', f'{current_path}'])
+                else:
+                    continue
+            subprocess.run(['touch', f'{reg_file}'])
+
+        with open(reg_file, 'r+') as f:
```
awcrosby, line 218: when are the cases the `reg_file` exists with text versus not? I'm wondering if there a way to make it more clear which code path is being followed and why

galaxy_importer/ansible_test/builders/pulp_build.py
```diff
@@ -15,7 +15,271 @@
...
+    def cleanup(self):
+        """Clean up temporary data structures and files"""
+        self._delete_container_repository(self.container_repository_href)
+        self._delete_container_distribution(self.container_distribution_href)
+        self._delete_artifact(self.dockerfile_artifact_href)
+        self._delete_artifact(self.collection_artifact_href)
+        os.remove(self.dockerfile)
+        os.remove(self.collection_file)
+
+    def _delete_container_repository(self, repository_href):
```
awcrosby, line 251: may be able to combine all 3 `_delete_*()` methods, passing in the url, checking for an "ok" status code via `requests.codes.ok` and in the exception saying which url caused the error - I am assuming that gives some context on what is attempting to be deleted.

PR #63, opened 2020-05-08
Parameterize openshift container sizing to fix scheduling issues

awcrosby:
Works https://github.com/ansible/galaxy_ng/issues/122

This PR adds parameters to openshift ansible-test container resources.

This allows for config to address container scheduling delays in CI and QA (i.e. lower `requests` and lower `limits` `cpu`) The lower cpu does not do a regression on the recent OOM issue https://github.com/ansible/galaxy_ng/issues/64.

### DIFF
--- a/galaxy_importer/ansible_test/job_template.yaml
+++ b/galaxy_importer/ansible_test/job_template.yaml
@@ -15,10 +15,10 @@ spec:
         image: {image}
         resources:
           requests:
-            cpu: 500m
-            memory: 512Mi
+            cpu: {cpu_request}
+            memory: {memory_request}
           limits:
-            cpu: 1
-            memory: 1Gi
+            cpu: {cpu_limit}
+            memory: {memory_limit}
       restartPolicy: Never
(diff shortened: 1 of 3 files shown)


galaxy_importer/ansible_test/job_template.yaml
```diff
@@ -14,11 +14,8 @@ spec:
       - name: {job_name}
         image: {image}
         resources:
-          requests:
```
cutwater, line 17: I don't think removing `requests:` is a good idea. If container is started with `limits`, but no `requests`, the `requests` will be (not 100% sure, but according to [1]) set to match limit, which is `1Gi`.

Also it would be nice to configure such job parameters, that environment dependent too as galaxy-importer parameters. So that it can be tuned for a specific environment without having to release a new version of a `galaxy-importer`.

[1] https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/memory-default-namespace/

awcrosby: added parameters with requests

PR #67, opened 2020-05-22
Travis publish to pypi on git tag

awcrosby:


.travis.yml
```diff
@@ -21,3 +21,13 @@ install:
 script:
   - make test/flake8
   - make test
+
+deploy:
+  provider: pypi
```
newswangerd, line 26: Might need to add a `distributions` key. I think by default it just builds a tarball, but it looks like you also provide a wheel.

https://github.com/ansible/galaxy_ng/blob/master/.travis.yml#L45

awcrosby: Yes, thanks, my pre-release test missed the wheel, added it in.

PR #68, opened 2020-06-03
Move from tarfile lib to tar in shell

awcrosby:


galaxy_importer/collection.py
```diff
@@ -57,8 +58,20 @@ def _import_collection(file, filename, logger, cfg):
...
-            pkg_tar.extractall(extract_dir)
+        os.makedirs(extract_dir)
+
+        cmd = [
+            'tar',
+            f'--directory={extract_dir}',
+            '-xf',
+            file.name,
+        ]
+        proc = Popen(cmd, cwd=PULP_ARTIFACT_DIR)
```
cutwater, line 69: You could simply use `subprocess.run()` to run command and return its exit code.

setup.cfg
```diff
@@ -41,5 +41,5 @@ galaxy_importer =
     ansible_test/build_template.yaml
 
 [coverage:report]
-fail_under = 93.7
+fail_under = 94.2
```
alikins, line 44: Unrelated change?

galaxy_importer/collection.py
```diff
@@ -57,8 +58,20 @@ def _import_collection(file, filename, logger, cfg):
...
+        return_code = proc.wait()
+        if return_code != 0:
+            raise exc.ImporterError('A tar error occurred in {}, returncode={}'.format(
+                ' '.join(cmd), return_code))
```
alikins, line 73: I would consider adding some error logging here with the return code / stdout / stderr / cmd.

That is logging to at least the "regular" app logger, and... maybe to the "import" logger that is used to collect and display import process.

Just for troubleshooting purposes and to be able to verify that any failures aren't exploit atttempts.

alikins: (depending on what all the SubprocessError includes, just logging it's str/repr may be enough. I'd have to try it and see)

galaxy_importer/collection.py
awcrosby: Updated

setup.cfg
awcrosby: I usually bump this up in same PR if coverage increases, keeps it from creeping back down

galaxy_importer/collection.py
awcrosby: Added more info in logging here, passing to display in import process

awcrosby:
@alikins @cutwater also made update to support being called by pulp_ansible, community galaxy, and galaxy_importer.main cli:

- Accept file object and file path in `import_collection()`. `filepath` used to untar the collection tarball. `file` fileobject only used for running ansible-test under pulp infrastructure utilizing the artifact file object.

galaxy_importer/collection.py
```diff
@@ -77,6 +84,24 @@ def _import_collection(file, filename, logger, cfg):
...
+        'tar',
+        f'--directory={extract_dir}',
+        '-xf',
+        file_name,
+    ]
+    try:
+        proc = run(args, cwd=cwd, stderr=PIPE)
+    except (FileNotFoundError, SubprocessError) as e:
+        raise exc.ImporterError(f'Error in tar extract subprocess: {str(e)}')
+    if proc.returncode != 0:
```
cutwater, line 100: Minor: You could use `run` with `check=True` argument. In this case `run` raises an exception in case of non zero exit code.

galaxy_importer/main.py
```diff
@@ -90,15 +90,18 @@ def call_importer(filepath, cfg):
...
-            logging.error('Unexpected error occurred:', exc_info=True)
-            return None
+    try:
+        data = collection.import_collection(
+            file=None,
+            filepath=filepath,
+            filename=filename,
+            cfg=cfg)
+    except ImporterError as e:
+        logging.error(f'The import failed for the following reason: {str(e)}')
```
alikins, line 100: Suggestion: Include the filename/filepath in the log message explicitly (in addition to anything that may be included in the str(e))'

As far as I can tell, the places where ImprotError is raised in the extract path show the str() of either FileNotFoundError or SubprocessError, which should usually include the name of the file being included.  But if there is an ImportError raised elsewhere, I would suggest going ahead and logging the filename/filepath contextual info as well. Some of the places ImportError derived classes are raised don't include much info.

galaxy_importer/collection.py
awcrosby: updated

galaxy_importer/main.py
```diff
@@ -90,15 +90,18 @@ def call_importer(filepath, cfg):
...
+        return None
+    except Exception:
+        logging.error('Unexpected error occurred:', exc_info=True)
```
alikins, line 103: You can setup a module scope logger in this module and use it here. Just using logging.error will use the root / unnamed logger which makes it harder tell where it is coming from.  For humans and for log processing tools. 

ie,
``` python
log = logging.getLogger(__name__)
# or if __name__ ends up __main__ (not sure what happens when entrypoints are used), then an
# explicit
# log = logging.getLogger('galaxy_importer.main')
...
     except Exception:
         log.exception(''Unexpected error occurred:')
```

log.exception() is equiv to log.error('some message', exc_info=True)

galaxy_importer/collection.py
```diff
@@ -77,6 +84,23 @@ def _import_collection(file, filename, logger, cfg):
...
+def _extract_tar_shell(filepath, extract_dir):
+    cwd = os.path.dirname(os.path.abspath(filepath))
+    file_name = os.path.basename(filepath)
+    args = [
+        'tar',
+        f'--directory={extract_dir}',
+        '-xf',
+        file_name,
+    ]
+    try:
```
alikins, line 96: I'd suggest doing the try/excepts in _import_collection  so _extract_tar_shell stays semi generic.

Handling the exceptions in _import_collection also gives you more context to construct exceptions and error messages.

galaxy_importer/collection.py
```diff
@@ -41,7 +42,7 @@
     namedtuple("CollectionFilename", ["namespace", "name", "version"])
 
 
-def import_collection(file, filename=None, logger=None, cfg=None):
+def import_collection(file=None, filepath=None, filename=None, logger=None, cfg=None):
```
alikins, line 45: Feels odd having all the args be optional. Plus 'file', 'filepath', 'filename'. I'm guessing this is to maintain API compat since this is the main "public" API?

PR #69, opened 2020-06-03
Add config option for 'run_ansible_lint'

alikins:


galaxy_importer/collection.py
```diff
@@ -220,6 +222,19 @@ def _build_contents_blob(self):
 
     def _build_docs_blob(self):
         """Build importer result docs_blob from collection documentation."""
+
+        # return an empty DocsBlob if run_ansible_doc=False
+        # if not self.cfg.run_ansible_doc:
```
awcrosby, line 227: not sure if this comment line is meant to stick around

chouseknecht:
@alikins 

Please add change log snippet and an update to the README example config. Thanks!

galaxy_importer/collection.py
cutwater: +1, please don't leave commented unused blocks of code.

awcrosby:
Looks like it still needs the README example config updated with the 2 config options, otherwise looks good

PR #74, opened 2020-06-19
Build and run ansible-test via local images

bmclaughlin:
Enhancement to build and run ansible-test via local images, using docker, for process isolation.

To test:
See the [End User Install](https://github.com/ansible/galaxy_ng/wiki/End-User-Installation) instructions and point the `galaxy-importer` to this branch to test against Automation Hub:

Issues:
    ansible/galaxy_ng#47
    ansible/galaxy_ng#81

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,9 +43,13 @@ Example configuration file:
 LOG_LEVEL_MAIN = INFO
 RUN_FLAKE8 = False
 RUN_ANSIBLE_TEST = False
-INFRA_LOCAL_IMAGE = False
+ANSIBLE_TEST_LOCAL_IMAGE = False
 INFRA_OSD = False
 ```
+
+- `RUN_ANSIBLE_TEST` - Set to `True` to run `ansible-test` during collection import. Defaults to `False`.
+
+- `ANSIBLE_TEST_LOCAL_IMAGE` - Set to `True` to run `ansible-test` sandboxed within a container image. Requires installation of Docker to run the container. Defaults to `False`.
 
 ### Issues and Process
 
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ Example configuration file:
 LOG_LEVEL_MAIN = INFO
 RUN_FLAKE8 = False
 RUN_ANSIBLE_TEST = False
-INFRA_PULP = False
+INFRA_LOCAL_IMAGE = False
 INFRA_OSD = False
 ```
 
--- a/galaxy_importer/ansible_test/builders/local_image_build.py
+++ b/galaxy_importer/ansible_test/builders/local_image_build.py
@@ -1,4 +1,4 @@
-# (c) 2012-2019, Ansible by Red Hat
+# (c) 2012-2020, Ansible by Red Hat
 #
 # This file is part of Ansible Galaxy
 #
@@ -15,8 +15,7 @@
 # You should have received a copy of the Apache License
 # along with Galaxy.  If not, see <http://www.apache.org/licenses/>.
 
-from galaxy_importer.ansible_test.builders import pulp_build
 
-
-def test_build_image_with_artifact():
-    assert pulp_build.build_image_with_artifact() == ''
+def build_image_with_artifact():
+    """Use docker/podman to build ansible-test image with artifact inside."""
+    return ''
--- a/galaxy_importer/ansible_test/builders/local_image_build.py
+++ b/galaxy_importer/ansible_test/builders/local_image_build.py
@@ -15,7 +15,91 @@
 # You should have received a copy of the Apache License
 # along with Galaxy.  If not, see <http://www.apache.org/licenses/>.
 
+import logging
+import os
+import requests
+import tempfile
 
-def build_image_with_artifact():
+from galaxy_importer import config
+from galaxy_importer import exceptions
+from shutil import copy
+from subprocess import Popen, PIPE, STDOUT, run, CalledProcessError
+
+config_data = config.ConfigFile.load()
+cfg = config.Config(config_data=config_data)
+default_logger = logging.getLogger(__name__)
+
+
+class Build(object):
     """Use docker/podman to build ansible-test image with artifact inside."""
-    return ''
+    def __init__(self, filepath, collection_name, logger=default_logger):
+        self.filepath = filepath
+        self.log = logger or default_logger
+        self.image = ''
+        self.working_dir = tempfile.TemporaryDirectory()
+
+    def build_image(self):
+        self.log.info('Building Dockerfile')
+        Build._build_dockerfile(self.working_dir.name)
+        Build._copy_collection_file(
+            dir=self.working_dir.name,
+            filepath=self.filepath
+        )
+
+        self.log.info('Building image...')
+        self.image = Build._build_image_with_artifact(dir=self.working_dir.name)
+        return self.image
+
+    def cleanup(self):
+        self.log.info('Removing temporary files, image and container')
+        self.working_dir.cleanup()
+
+        cmd = ['docker', 'image', 'rm', '-f', self.image]
+        try:
+            run(cmd)
+        except CalledProcessError as e:
+            raise exceptions.AnsibleTestError(
+                'An exception occurred in: {}, message={}'.format(' '.join(cmd), e.msg))
+
+    def _build_dockerfile(dir):
+        url = 'https://raw.githubusercontent.com/ansible/galaxy-importer \
+            /master/docker/ansible-test/Dockerfile'
+        file_location = os.path.join(dir, 'Dockerfile')
+        with requests.get(url, allow_redirects=True) as r:
+            with open(file_location, 'wb') as f:
+                f.write(r.content)
+        with open(file_location, 'r+') as f:
+            lines = f.readlines()
+            for index, line in enumerate(lines):
+                if 'ENV HOME' in line:
+                    # TODO move RUN command to base image
+                    lines.insert(index - 1, '\nRUN chown -R user1:root /archive\n')
+                    lines.insert(index - 1, '\nCOPY archive.tar.gz /archive/archive.tar.gz\n')
+                    break
+            f.seek(0)
+            f.writelines(lines)
+
+    def _build_image_with_artifact(dir):
+        cmd = ['docker', 'build', '.', '--quiet']
+        proc = Popen(
+            cmd,
+            cwd=dir,
+            stdout=PIPE,
+            stderr=STDOUT,
+            encoding='utf-8',
+        )
+
+        result = ''
+        for line in proc.stdout:
+            result = line.strip()
+
+        return_code = proc.wait()
+        if return_code != 0:
+            raise exceptions.AnsibleTestError(
+                'An exception occurred in {}, returncode={}'
+                .format(' '.join(cmd), return_code))
+        return result.split(':')[-1]
+
+    def _copy_collection_file(dir, filepath):
+        path = os.path.join(dir, 'archive.tar.gz')
+        copy(filepath, path)
--- a/galaxy_importer/ansible_test/runners/__init__.py
+++ b/galaxy_importer/ansible_test/runners/__init__.py
@@ -31,7 +31,7 @@ def get_runner(cfg):
     if not cfg.run_ansible_test:
         return None
 
-    if not cfg.infra_pulp:
+    if not cfg.infra_local_image:
         return LocalAnsibleTestRunner
 
     if cfg.infra_osd:
--- a/galaxy_importer/ansible_test/runners/__init__.py
+++ b/galaxy_importer/ansible_test/runners/__init__.py
@@ -31,10 +31,10 @@ def get_runner(cfg):
     if not cfg.run_ansible_test:
         return None
 
-    if not cfg.infra_local_image:
-        return LocalAnsibleTestRunner
-
     if cfg.infra_osd:
         return OpenshiftJobTestRunner
-    else:
+
+    if cfg.ansible_test_local_image:
         return LocalImageTestRunner
+
+    return LocalAnsibleTestRunner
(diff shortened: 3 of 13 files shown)


README.md
```diff
@@ -43,7 +43,7 @@ Example configuration file:
 LOG_LEVEL_MAIN = INFO
 RUN_FLAKE8 = False
 RUN_ANSIBLE_TEST = False
-INFRA_PULP = False
+INFRA_LOCAL_IMAGE = False
```
chouseknecht, line 46: Add `INFRA_LOCAL_DOCKER = False`

Below the example config file, please add something like the following to explain the `ansible-test` options:

* RUN_ANSIBLE_TEST - Set to `True` to run `ansible-test` during collection import. If `True`, requires Ansible to be installed, unless container image execution is enabled.
* INFRA_LOCAL_IMAGE - Set to `True` to run `ansible-test` sandboxed within a container image. Defaults to  False. Requires installation of either Podman or Docker to run the container.
* INFRA_LOCAL_DOCKER - Set to `True` to run the `ansible-test` container image via Docker; otherwise, Podman will be used.

awcrosby: I thought we were just going to pick one local execution, docker or podman, if so i vote docker.

chouseknecht: What's wrong with offering both? Plus, Podman doesn't require a daemon.

awcrosby: sounds good

README.md
```diff
@@ -43,10 +43,17 @@ Example configuration file:
...
+INFRA_LOCAL_DOCKER = False
 INFRA_OSD = False
 ```
 
+- `RUN_ANSIBLE_TEST` - Set to `True` to run `ansible-test` during collection import. If `True`, requires Ansible to be installed, unless container image execution is enabled. Defaults to `False`.
+
+- `INFRA_LOCAL_IMAGE` - Set to `True` to run `ansible-test` sandboxed within a container image. Requires installation of either Podman or Docker to run the container. Defaults to `False`.
```
awcrosby, line 53: I think a config value that denotes when to run ansible-test in local container is good. Name suggestion: `ansible_test_local_image` and `local_image_docker` (originally infra* was denoting things like if backend was openshift or not, but that convention might not be very useful after all)

galaxy_importer/ansible_test/builders/local_image_build.py
```diff
@@ -0,0 +1,93 @@
...
+default_logger = logging.getLogger(__name__)
+
+
+class Build(object):
+    """Use docker/podman to build ansible-test image with artifact inside."""
+    def __init__(self, collection_file, collection_name, logger):
+        self.collection_file = collection_file
+        self.container_engine = Build.get_container_engine()
+        self.log = logger or default_logger
+        self.collection_name = collection_name
```
awcrosby, line 40: i believe this is the collection version name rather than the collection name

galaxy_importer/ansible_test/builders/local_image_build.py
```diff
@@ -0,0 +1,93 @@
...
+        self.tag = f'galaxy-importer:{self.collection_name}'
+        self.working_dir = f'/tmp/galaxy-importer-{str(uuid.uuid4())}/'
```
awcrosby, line 42: should probably use `tempfile.TemporaryDirectory` if possible

galaxy_importer/ansible_test/builders/local_image_build.py
```diff
@@ -0,0 +1,93 @@
...
+
+    @staticmethod
+    def get_container_engine():
+        if cfg.infra_local_docker is True:
+            return 'docker'
+        else:
+            return 'podman'
+
+    def build_dockerfile(self):
+        self.log.info('Building dockerfile')
```
awcrosby, line 52: would be good to have log entry for users at other step(s) - especially the docker build as that can take a while.

possibly say 'Building Containerfile' for more neutral language?, as I think podman uses that (also since we are defaulting to podman)

galaxy_importer/ansible_test/builders/local_image_build.py
```diff
@@ -0,0 +1,93 @@
...
+        url = 'https://raw.githubusercontent.com/ansible/galaxy-importer/master/docker/ansible-test/Dockerfile'
```
awcrosby, line 53: this may be a good reason in the future for us to bring the dockerfile inside the `galaxy-importer` package (and update where openshift looks for it), at that point we could avoid this download which will happen on every import

galaxy_importer/ansible_test/builders/local_image_build.py
```diff
@@ -0,0 +1,93 @@
...
+        return self.tag
+
+    def cleanup(self):
+        rmtree(self.working_dir)
+
+    def copy_collection_file(self):
+        path = os.path.join(self.working_dir, 'archive.tar.gz')
+
+        try:
+            url = Build._get_pulp_archive_url(self.pulp_artifact_file)
```
awcrosby, line 77: this is handling 2 scenarios: called via `galaxy_importer.main` and called via `pulp_ansible` w/ external storage. we'll also need to handle the case where pulp is using local storage.

I think https://github.com/ansible/galaxy-importer/pull/71 will help here - it downloads the archive if needed, that way we won't need to download the file a 2nd time here. The filepath can be passed here and this probably won't need to know about pulp implementation.

galaxy_importer/ansible_test/builders/local_image_build.py
```diff
@@ -0,0 +1,93 @@
...
+            with requests.get(url, stream=True) as r:
+                with open(path, 'wb') as f:
+                    copyfileobj(r.raw, f)
+        except AttributeError:
+            spath = os.path.join(
+                os.path.dirname(constants.ROOT_DIR.rstrip('galaxy_importer')),
```
awcrosby, line 83: this assumes the archive is in the path of the package, which won't be the case. the file passed in to the importer is accessible here as `self.collection_file`, but if we can use filepath noted above, that would be best.

galaxy_importer/ansible_test/runners/__init__.py
```diff
@@ -31,7 +31,7 @@ def get_runner(cfg):
     if not cfg.run_ansible_test:
         return None
 
-    if not cfg.infra_pulp:
+    if not cfg.infra_local_image:
```
awcrosby, line 34: since we now have a config value for this scenario, we can say this (or whatever the final config param name is)
```
if cfg.ansible_test_local_image:
    return LocalImageTestRunner
```

galaxy_importer/ansible_test/builders/local_image_build.py
bmclaughlin: I was being consistent with https://github.com/ansible/galaxy-importer/blob/master/galaxy_importer/ansible_test/runners/local_ansible_test.py#L45.  Should I update both?

README.md
chouseknecht: Sounds good to me.

galaxy_importer/ansible_test/runners/__init__.py
```diff
@@ -31,10 +31,11 @@ def get_runner(cfg):
     if not cfg.run_ansible_test:
         return None
 
-    if not cfg.infra_pulp:
+    if not cfg.ansible_test_local_image:
```
awcrosby, line 34: Since this value defaults to false, if it is not in the config it will win, I think we want this:
```
    if cfg.infra_osd:
        return OpenshiftJobTestRunner

    if cfg.ansible_test_local_image:
        return LocalImageTestRunner

    return LocalAnsibleTestRunner
```

galaxy_importer/constants.py
```diff
@@ -16,9 +16,10 @@
 # along with Galaxy.  If not, see <http://www.apache.org/licenses/>.
 
 import enum
+import os
 import re
 
-
+ROOT_DIR = os.path.dirname(os.path.realpath(__file__))
```
awcrosby, line 22: This can be removed as no longer used

tests/test_builder_local_image.py
```diff
@@ -15,8 +15,19 @@
...
+@pytest.fixture
+def build():
+    return local_image_build.Build('/file/path/to/archive.tar.gz', 'name')
+
+
+def test_build_image_with_artifact(build, mocker):
+    mocker.patch.object(subprocess, 'run')
+    build.build_image_with_artifact()
+    assert subprocess.run.called
+    assert build.build_image_with_artifact() == 'galaxy-importer:name'
```
awcrosby, line 33: can use result of call above here

tests/test_runner.py
```diff
@@ -51,20 +51,20 @@ def test_get_runner_ansible_test_local(temp_config_file):
...
-                'INFRA_PULP = True')
+                'ANSIBLE_TEST_LOCAL_IMAGE = True')
         f.flush()
         config_data = config.ConfigFile.load()
         cfg = config.Config(config_data=config_data)
         assert runners.get_runner(cfg) == runners.LocalImageTestRunner
 
 
-def test_get_runner_pulp_and_osd(temp_config_file):
+def test_get_runner_local_image_and_osd(temp_config_file):
```
awcrosby, line 64: Having both `ANSIBLE_TEST_LOCAL_IMAGE` and `INFRA_OSD` as `True` conflict in terms of picking a Runner, so this test changes nature a bit to see which config option wins, probably should now be named `test_osd_config_wins_over_local_image()`

galaxy_importer/ansible_test/runners/local_image.py
```diff
@@ -15,11 +15,45 @@
...
-        pass
+        build = Build(
+            self.filepath,
+            f'{self.metadata.namespace}-{self.metadata.name}-{self.metadata.version}',
+            self.log)
+
+        build.build_dockerfile()
+        build.copy_collection_file()
+
+        tag = build.build_image_with_artifact()
```
awcrosby, line 36: minor... ideally, these could be encapsulated into a a single method that returns the tag, so runner wouldn't know about inner workings, something like`.build_image()`, and things like `copy_collection_file()` can be an internal class methods

README.md
awcrosby: We should remove `If 'True', requires Ansible to be installed, unless container image execution is enabled.` - as ansible is required by the package in general for ansible-doc and ansible-lint

galaxy_importer/ansible_test/builders/local_image_build.py
awcrosby: Ah yes, staying consistent is good, and probably not worth changing the other.

tests/test_builder_local_image.py
```diff
@@ -0,0 +1,66 @@
...
+    with open(tmp_file, 'w') as f:
+        f.write('file contents go here')
+        f.flush()
+        build = builders.local_image_build.Build(
+            filepath=tmp_file,
+            collection_name='namespace-name-version'
+        )
+        mocker.patch.object(build, '_build_dockerfile')
+        mocker.patch.object(build, '_copy_collection_file')
+        mocker.patch.object(build, '_build_image_with_artifact')
```
awcrosby, line 50: These end up not mocking the methods called, see [gist](https://gist.github.com/awcrosby/f9144e83b6f266f3c271c88a970da468
)
`_build_dockerfile()` should probably have it's own test method.

galaxy_importer/ansible_test/builders/local_image_build.py
```diff
@@ -0,0 +1,92 @@
...
+        url = 'https://raw.githubusercontent.com/ansible/galaxy-importer/master/docker/ansible-test/Dockerfile'
+        file_location = os.path.join(dir, 'Dockerfile')
+        with requests.get(url, allow_redirects=True) as r:
+            with open(file_location, 'wb') as f:
+                f.write(r.content)
+        with open(file_location, 'r+') as f:
+            lines = f.readlines()
+            for index, line in enumerate(lines):
+                if 'ENV HOME' in line:
+                    lines.insert(index - 1, '\nCOPY archive.tar.gz /archive/archive.tar.gz\nRUN chown -R user1:root /archive\n')
```
awcrosby, line 77: We can probably add `RUN chown -R user1:root /archive` to the base image used by OSD and local image execution, but this can be done later

awcrosby: actually, could you put a TODO here as a reminder of ^?

tests/test_builder_local_image.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+def test_build_dockerfile(mocker):
+    dir = tempfile.TemporaryDirectory()
+    Dockerfile_updated_correctly = False
+
+    Build._build_dockerfile(dir.name)
+
+    with open(f'{dir.name}/Dockerfile') as f:
+        if 'COPY archive.tar.gz /archive/archive.tar.gz' in f.read():
```
awcrosby, line 72: you can use `assert 'COPY archive.tar.gz /archive/archive.tar.gz' in f.read()` rather than the `Dockerfile_updated_correctly` var

tests/test_builder_local_image.py
```diff
@@ -0,0 +1,106 @@
...
+            Dockerfile_updated_correctly = True
+
+    assert Dockerfile_updated_correctly
+    dir.cleanup()
+
+
+@mock.patch('galaxy_importer.ansible_test.builders.local_image_build.Popen')
+def test_build_image_with_artifact(mocked_popen, mocker):
+    dir = tempfile.TemporaryDirectory()
```
awcrosby, line 81: missing a `dir.cleanup()`

these calls can also be done in a context manager: `with tempfile.TemporaryDirectory as dir:` which is cleaned up automatically

tests/test_builder_local_image.py
```diff
@@ -0,0 +1,106 @@
...
+    mocked_popen.return_value.stdout = ['sha256:1234', 'sha256:5678']
+    mocked_popen.return_value.wait.return_value = 0
+    result = Build._build_image_with_artifact(dir=dir.name)
+    assert mocked_popen.called
+    assert '5678' in result
+
+
+@mock.patch('galaxy_importer.ansible_test.builders.local_image_build.Popen')
+def test_build_image_with_artifact_exception(mocked_popen, mocker):
+    dir = tempfile.TemporaryDirectory()
```
awcrosby, line 91: also cleanup here

bmclaughlin:
@awcrosby, thanks for the suggestions on the tests.  Updated!

PR #79, opened 2020-07-16
Require collection tags include 1+ tags from the required tag list

bmclaughlin:
The documentation will need to be updated afterwards [1].  

Issue: https://github.com/ansible/galaxy_ng/issues/255
[1] https://docs.ansible.com/ansible/latest/dev_guide/collections_galaxy_meta.html#collections-galaxy-meta

### DIFF
--- a/galaxy_importer/schema.py
+++ b/galaxy_importer/schema.py
@@ -21,10 +21,24 @@ import re
 import attr
 import semantic_version
 
+from galaxy_importer import config
 from galaxy_importer import constants
 from galaxy_importer.utils.spdx_licenses import is_valid_license_id
 
 SHA1_LEN = 40
+REQUIRED_TAG_LIST = [
+    'application',
+    'cloud',
+    'database',
+    'infrastructure',
+    'linux',
+    'monitoring',
+    'networking',
+    'security',
+    'storage',
+    'tools',
+    'windows',
+]
 
 
 def convert_none_to_empty_dict(val):
@@ -189,7 +203,7 @@ class CollectionInfo(object):
 
     @tags.validator
     def _check_tags(self, attribute, value):
-        """Check max tags and check against tag regular expression."""
+        """Check max tags and check against both tag regular expression and required tag list."""
         if value is not None and len(value) > constants.MAX_TAGS_COUNT:
             self.value_error(
                 f"Expecting no more than {constants.MAX_TAGS_COUNT} tags "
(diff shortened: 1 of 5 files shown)


galaxy_importer/schema.py
```diff
@@ -189,14 +202,16 @@ def _check_dependencies_format(self, attribute, dependencies):
...
-        """Check max tags and check against tag regular expression."""
+        """Check max tags and check against both tag regular expression and required tag list."""
         if value is not None and len(value) > constants.MAX_TAGS_COUNT:
             self.value_error(
                 f"Expecting no more than {constants.MAX_TAGS_COUNT} tags "
                 "in metadata")
         for tag in value:
             if not re.match(constants.NAME_REGEXP, tag):
                 self.value_error(f"'tag' has invalid format: {tag}")
+        if not any(tag in REQUIRED_TAG_LIST for tag in value):
```
awcrosby, line 213: This check will probably need to be configurable, defaulted to off. We want this check to occur in Community Galaxy and Automation Hub, but this would break most users of `pulp_ansible` independent of `galaxy` and `galaxy_ng`.

tests/integration/run_importer.sh
```diff
@@ -6,6 +6,7 @@ TMPDIR=$(mktemp -d -t galaxy-importer-XXXXXXXX)
 python3 -m venv $TMPDIR/venv
 source $TMPDIR/venv/bin/activate
 pip install .
+cp -R skeleton $TMPDIR/skeleton
```
awcrosby, line 9: we probably will not need integration test edits when this feature is defaulted to off

galaxy_importer/schema.py
```diff
@@ -189,14 +202,16 @@ def _check_dependencies_format(self, attribute, dependencies):
...
+            self.value_error('Missing required tag.')
```
awcrosby, line 214: We probably want this as clear as possible, as many users will hit this once it's turned on in Galaxy. Maybe: `f'At least one tag required from tag list: {", ".join(REQUIRED_TAG_LIST)}'`

bmclaughlin: Agreed, I was going back and forth on how much information to include here.

tests/unit/test_schema_collection_info.py
```diff
@@ -29,7 +31,7 @@ def collection_info():
         'license': ['MIT'],
         'readme': 'README.md',
         'authors': ['Bob Smith <b.smith@acme.com>'],
-        'tags': ['testcases'],
+        'tags': ['application'],
```
awcrosby, line 34: Since this check is defaulted to false, do existing tests need edits?

tests/unit/test_schema_collection_info.py
```diff
@@ -141,15 +143,40 @@ def test_invalid_tags(collection_info, invalid_tags):
...
+def temp_config_file():
+    try:
+        dir = os.path.dirname(os.path.dirname(os.path.dirname(__file__)))
+        config_file = os.path.join(dir, 'galaxy_importer', 'galaxy-importer.cfg')
+        yield config_file
+    finally:
+        os.remove(config_file)
+
+
+def test_required_tag(collection_info, temp_config_file):
```
awcrosby, line 169: probably need a test with config param set to true and it passes

PR #82, opened 2020-07-29
Fix Travis CI error when running against non master branch

cutwater:


cutwater:
```The command "PR_FILES=$(git diff --name-only $TRAVIS_BRANCH...HEAD)" failed and exited with 128 during .```

PR #92, opened 2020-08-11
Import and lint execution environment file

bmclaughlin:
For execution environment documentation see documents linked in this issue's Story:
https://issues.redhat.com/browse/AAH-3

Issue: https://issues.redhat.com/browse/AAH-9


galaxy_importer/collection.py
```diff
@@ -261,10 +261,29 @@ def _build_docs_blob(self):
...
+                linting_result = markup_utils.lint_execution_environment_file(ex_env_path)
+                for line in linting_result:
+                    self.log.warning(line)
+            except Exception as e:
+                self.log.error(f'Error during linting of execution environment metadata: {e}')
+
+            try:
+                execution_environment = markup_utils.get_execution_environment(ex_env_path)
+            except Exception as e:
+                self.log.error(f'Error during parsing of execution environment metadata: {e}')
```
awcrosby, line 280: I recommend pulling this out into something like `CollectionLoader._process_execution_environment()` that returns `execution_environment`. It will also make it easier to unit test without all the extra mocks.

galaxy_importer/utils/markup.py
```diff
@@ -77,6 +80,19 @@ def get_html(doc_file):
     return None
 
 
+def get_execution_environment(env_file):
```
awcrosby, line 83: wondering if we should have these in a `utils/yaml.py` module, since yaml isn't markup

galaxy_importer/collection.py
bmclaughlin: Thanks!

galaxy_importer/utils/markup.py
bmclaughlin: Makes perfect sense.  Done!

tests/unit/test_collection.py
```diff
@@ -263,6 +267,59 @@ def test_build_docs_blob_no_readme(get_readme_doc_file):
...
+
+
+EX_ENV_YAML = """version: 1
+dependencies:
+  python:
+    - somepkg==1.3
+    - otherpkg>=3.0
+  files:
+    - "/usr/bin/oc"
+    - "/usr/lib/libssl.so.1"
```
AlanCoding, line 286: I'm sorry about this schema, we've done a bad job of communicating the large amount of changes over to the issues you had open.

If you look at the README.md in https://github.com/ansible/ansible-builder, you can see the most up-to-date spec of how we will read data from collections. `files:` was never a part of what was actually being implemented, and there are other details about how the python/system deps are specified too.

galaxy_importer/utils/execution_environment.py
```diff
@@ -0,0 +1,88 @@
...
+        logger.info('Loading python dependencies')
+        python_contents = _load_python(python_path)
+        ex_env = _write_to_ee(ex_env, 'python', python_contents)
+    if os.path.exists(system_path):
+        logger.info('Loading system dependencies')
+        system_contents = _load_list(system_path)
+        ex_env = _write_to_ee(ex_env, 'system', system_contents)
+
+    if 'dependencies' not in ex_env.keys():
+        logger.info('No execution environment data found.')
```
awcrosby, line 56: you can put this towards the top and exit early before you read the dependencies, then you can get rid of the `if 'dependencies' in ex_env.keys():` and its `else`. As-is we are saying no data found but we are still loading and saving the data.

tests/unit/test_execution_environment_utils.py
```diff
@@ -0,0 +1,292 @@
...
+
+
+EE_YAML = """---
+version: 1
+dependencies:
+  galaxy: requirements.yml
+  python: requirements.txt
+  system: bindep.txt
+
+additional_build_steps:
```
awcrosby, line 35: is the `additional_build_steps` expected to be included in the `docs_blob` with the contents of the files in `dependencies`, or is that not needed?

galaxy_importer/utils/execution_environment.py
```diff
@@ -0,0 +1,88 @@
...
+        ex_env = _load_yaml(ex_env_path, logger)
+
+    if 'dependencies' in ex_env.keys():
+        system_path = os.path.join(path, ex_env['dependencies']['system'])
+        galaxy_path = os.path.join(path, ex_env['dependencies']['galaxy'])
+        python_path = os.path.join(path, ex_env['dependencies']['python'])
+    else:
+        system_path = os.path.join(path, 'bindep.txt')
+        galaxy_path = os.path.join(path, 'requirements.yml')
+        python_path = os.path.join(path, 'requirements.txt')
```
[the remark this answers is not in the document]
bmclaughlin: It's not a requirement but is the current common location where those files are being included in collections at this time.   I think the intention behind checking the current common location is to make this an easier feature for users to take advantage of.  I'll take it out to avoid any confusion.

tests/unit/test_execution_environment_utils.py
bmclaughlin: Yes it's expected to be included.

galaxy_importer/utils/execution_environment.py
```diff
@@ -0,0 +1,79 @@
...
+    ex_env = {}
+    ex_env_path = os.path.join(path, 'meta', 'execution-environment.yml')
+
+    if os.path.exists(ex_env_path):
+        ex_env = _load_yaml(ex_env_path, logger)
+    else:
+        logger.info('No execution environment data found.')
+        return ex_env
+
+    if os.path.exists(os.path.join(path, ex_env['dependencies']['galaxy'])):
```
awcrosby, line 34: Are we checking if these keys exist or handling KeyError anywhere? in case any of these keys are missing we might throw a stack trace to user

tests/unit/test_execution_environment_utils.py
awcrosby: ah yes i see now

galaxy_importer/utils/execution_environment.py
```diff
@@ -0,0 +1,88 @@
...
+    if 'dependencies' in ex_env.keys():
+        system_path = os.path.join(path, ex_env['dependencies']['system'])
+        galaxy_path = os.path.join(path, ex_env['dependencies']['galaxy'])
+        python_path = os.path.join(path, ex_env['dependencies']['python'])
+    else:
+        system_path = os.path.join(path, 'bindep.txt')
+        galaxy_path = os.path.join(path, 'requirements.yml')
+        python_path = os.path.join(path, 'requirements.txt')
+
+    if os.path.exists(galaxy_path):
```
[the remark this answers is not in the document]
awcrosby: might be good to do for the `galaxy`, `python`, `system` paths also

tests/unit/test_yaml_utils.py
```diff
@@ -0,0 +1,81 @@
...
+  python: requirements.txt
+  system: bindep.txt
+
+additional_build_steps:
+  prepend: |
+    RUN whoami
+    RUN cat /etc/os-release
+  append:
+    - RUN echo This is a post-install command!
+    - RUN ls -la /etc
```
AlanCoding, line 48: the additional_build_steps wouldn't be supported inside of collections. Collections could do all kinds of things which would be incompatible with other collections, so that's only left as something the user can add with writing their execution environment definition. It shouldn't need to be processed by the galaxy importer at all.

tests/unit/test_execution_environment_utils.py
```diff
@@ -0,0 +1,292 @@
...
+ansible
+paramiko
+"""
+
+BINDEP_TXT = """
+gcc
+python3-devel
+libcurl-devel
+openssl-devel
+libxml2-devel
```
AlanCoding, line 97: minor nit - it's expected to have the bindep data on it like `libxml2-devel [platform:rpm]`, but at the stage of processing you're doing, it won't do anything with that, so it would get that entry but then return it as-is.

tests/unit/test_yaml_utils.py
bmclaughlin: Thanks for the clarification, I'll update my PR accordingly.

tests/unit/test_execution_environment_utils.py
```diff
@@ -0,0 +1,292 @@
...
+            'version': 'ee_req',
+            'type': 'git'
+        },
+    ]
+}
+
+
+REQUIREMENTS_TXT = """
+ansible
+paramiko
```
AlanCoding, line 89: these might not be great examples... because they are liable to be in the list of requirements we sanitize

https://github.com/ansible/ansible-builder/blob/devel/ansible_builder/requirements.py#L4

Whether or not you sanitize the list of requirements here or not is a very important and very big question. Since you're pulling in the python requirements parser library, that would indicate that you _are_ planning to do requirements sanitization, and that will duplicate more code from `ansible-builder` beyond the `introspect.py` script. I think you could get by without doing that sanitization. If you do pull that in, we will have two code bases with a lot of logic which will quickly start to diverge and it's going to get really hard to maintain.

tests/unit/test_execution_environment_utils.py
bmclaughlin: I didn't come across any `bindep.txt` files with the platform data.  I'll track down an example and update my tests.

galaxy_importer/utils/execution_environment.py
```diff
@@ -0,0 +1,116 @@
...
+    with open(path, 'r') as f:
+        pip_content = f.read()
+
+    pip_lines = []
+    for line in pip_content.split('\n'):
+        if line_is_empty(line):
+            continue
+        if '#' in line:
+            line = re.sub(r' *#.*\n?', '', line)
+        if line.startswith('-r') or line.startswith('--requirement'):
```
awcrosby, line 91: we should probably have tests with `-r` and `--requirement`

galaxy_importer/utils/execution_environment.py
```diff
@@ -0,0 +1,116 @@
...
+    ):
+        logger.info('Linting collection dependencies')
+        galaxy_contents = _load_yaml(
+            os.path.join(path, ex_env['dependencies']['galaxy']),
+            logger
+        )
+        logger.info('Loading collection dependencies')
+        ex_env = _write_to_ee(ex_env, 'galaxy', galaxy_contents)
+    else:
+        logger.warning('Galaxy dependencies file not found')
```
awcrosby, line 45: If we always expect there to be sections for `galaxy`, `python`, `system` and therefore files, then these warnings make sense. If it is OK that a user doesn't enter a `galaxy` section for example, we still warn them `file not found` - I think that could be confusing.

If a missing dependencies key should throw a warning, could check and exit early:
```
if 'dependencies' not in ex_env:
    logger.warning('No dependencies found')
    return ex_env
```

AlanCoding: Only python and system entries here make sense. The spec allows the _user's definition_ to give a Galaxy requirements file. On the collection level, dependencies are specified in `galaxy.yml`, not here. If anything, you should error if you find a `galaxy` entry here.

galaxy_importer/utils/execution_environment.py
```diff
@@ -0,0 +1,116 @@
...
+
+def process_execution_environment(path, logger):
+    ex_env = {}
+    ex_env_path = os.path.join(path, 'meta', 'execution-environment.yml')
+
+    if os.path.exists(ex_env_path):
+        ex_env = _load_yaml(ex_env_path, logger)
+    else:
+        logger.info('No execution environment data found.')
+        return ex_env
```
AlanCoding, line 32: Here this is an empty dict `{}`, to get parity with what `ansible-builder` is doing, you could return something like:

```json
{
  "dependencies": {
    "python": "requirements.txt",
    "system": "bindep.txt"
  }
}
```

...conditionally depending on whether each of those 2 files existed.

galaxy_importer/utils/execution_environment.py
AlanCoding: My original source for this is:

https://github.com/davidfischer/requirements-parser/blob/master/requirements/parser.py

They have lots of test cases there you can tap into, I have a cursory 1 or 2 file structures in ansible-builder.

galaxy_importer/utils/execution_environment.py
```diff
@@ -0,0 +1,105 @@
...
+    if 'dependencies' not in ex_env:
+        logger.warning('No dependencies found')
+        return ex_env
+
+    python_exists = os.path.exists(os.path.join(path, ex_env['dependencies']['python']))
+    system_exists = os.path.exists(os.path.join(path, ex_env['dependencies']['system']))
+
+    if 'python' in ex_env['dependencies'] and python_exists:
```
awcrosby, line 40: we are not handling the case where `python` is not in `ex_env['dependencies']` ... because of `ex_env['dependencies']['python']` this causes an exception:
```
version: 1
dependencies:
  system: bindep.txt
```

galaxy_importer/utils/execution_environment.py
```diff
@@ -0,0 +1,105 @@
...
+        logger.info('Loading python dependencies')
+        python_contents = _pip_file_data(os.path.join(path, ex_env['dependencies']['python']))
+        ex_env = _write_to_ee(ex_env, 'python', python_contents)
+    elif 'python' in ex_env['dependencies'] and not python_exists:
+        logger.warning('Python dependencies file not found')
+
+    if 'system' in ex_env['dependencies'] and system_exists:
```
awcrosby, line 47: ditto

galaxy_importer/utils/yaml.py
```diff
@@ -0,0 +1,34 @@
...
+
+from yamllint.config import YamlLintConfig
+from yamllint import linter
+
+
+def safe_load_file(file_path):
+    d = {}
+    with open(file_path) as fp:
+        d = yaml.safe_load(fp)
+    return d
```
awcrosby, line 28: For an empty file this will return `NoneType` and cause exception later, if you change this to `return d or {}` it should fix that

galaxy_importer/utils/execution_environment.py
```diff
@@ -0,0 +1,105 @@
...
+def process_execution_environment(path, logger):
+    ex_env = {}
+    ex_env_path = os.path.join(path, 'meta', 'execution-environment.yml')
+
+    if os.path.exists(ex_env_path):
+        ex_env = _load_yaml(ex_env_path, logger)
+    else:
+        logger.info('No execution environment data found.')
+        return ex_env
+    if 'dependencies' not in ex_env:
```
awcrosby, line 33: an empty dependencies will cause an exception, this should prevent that `if 'dependencies' not in ex_env or not ex_env['dependencies']:`

awcrosby (changes requested):

galaxy_importer/utils/execution_environment.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+def process_execution_environment(path, logger):
+    ex_env = {}
+    ex_env_path = os.path.join(path, 'meta', 'execution-environment.yml')
+
+    if os.path.exists(ex_env_path):
+        ex_env = _load_yaml(ex_env_path, logger)
+    else:
+        logger.info('No execution environment data found.')
```
awcrosby, line 31: very minor... the style on the other importer log messages is without a period

AlanCoding:
Looking at this now, I feel like it is still unlikely to have parity what what `ansible-builder` does, but I could be wrong.

I put up this:

https://github.com/ansible/ansible-builder/pull/97

I don't like that this duplicates code with what ansible-builder does, and I truly believe this is an unmaintainable pattern. My intent with this PR is that you can outright copy the `introspect.py` module, put it wherever you want in the source for the galaxy importer and only use that module for producing dependency data. Even if you copy and paste the module, that's still an improvement because we can also update it with a simple copy + paste. What we have now will diverge unless dedicated effort is constantly put into updating the different code bases that approach the problem differently.

awcrosby:
@AlanCoding I agree about the duplication / diverging / maintenance point, copy and paste or depending on ansible-builder is probably the best approach here

AlanCoding (approved):
Looks like relatively good test coverage

Looking over the interface points, nothing jumps out to me as a problem after the latest refactor, thanks for making those changes.

galaxy_importer/utils/execution_environment.py
```diff
@@ -0,0 +1,44 @@
...
+
+def process_execution_environment(path, logger):
+    deps = ansible_builder.introspect.process_collection(path)
+    ex_env = {}
+
+    if len(deps[0]) == 0 and len(deps[1]) == 0:
+        logger.info('No execution environment dependencies found')
+        return ex_env
+
+    if len(deps[0]) > 0:
```
awcrosby, line 29: you can use the truthiness of lists to say `if deps[0]:`

tests/unit/test_execution_environment_utils.py
```diff
@@ -0,0 +1,186 @@
...
+                ]
+            }
+        }
+    finally:
+        for f in files:
+            os.remove(f)
+        os.rmdir(os.path.join(dir, 'meta'))
+
+
+def test_process_execution_environment_no_python():
```
awcrosby, line 79: what is the resulting dict supposed to look like if `meta/execution-environment.yml` does not say there are python dependencies, but the python dependencies file `requirements.txt` exists (and vice versa for system deps)... can we have tests that cover this?

awcrosby (approved):
LGTM

PR #94, opened 2020-10-26
Make root temp and ansible local temp directories configurable

bmclaughlin:
Issue: https://issues.redhat.com/browse/AAH-91

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,7 @@ RUN_ANSIBLE_TEST = False
 ANSIBLE_TEST_LOCAL_IMAGE = False
 INFRA_OSD = False
 TMP_ROOT_DIR = None
+ANSIBLE_LOCAL_TMP = '~/.ansible/tmp'
 ```
 
 - `RUN_ANSIBLE_TEST` - Set to `True` to run `ansible-test` during collection import. Defaults to `False`.
--- a/README.md
+++ b/README.md
@@ -47,6 +47,7 @@ RUN_ANSIBLE_LINT = True
 RUN_ANSIBLE_TEST = False
 ANSIBLE_TEST_LOCAL_IMAGE = False
 INFRA_OSD = False
+TMP_ROOT_DIR = None
 ```
 
 - `RUN_ANSIBLE_TEST` - Set to `True` to run `ansible-test` during collection import. Defaults to `False`.
--- a/galaxy_importer/collection.py
+++ b/galaxy_importer/collection.py
@@ -153,6 +153,7 @@ class CollectionLoader(object):
                 path=self.path,
                 fq_collection_name='{}.{}'.format(self.metadata.namespace, self.metadata.name),
                 logger=self.log,
+                cfg=self.cfg,
             ).load()
 
         self.content_objs = list(self._load_contents())
--- a/galaxy_importer/collection.py
+++ b/galaxy_importer/collection.py
@@ -58,10 +58,7 @@ def import_collection(file, filename=None, logger=None, cfg=None):
 
 
 def _import_collection(file, filename, logger, cfg):
-    tmp_root_dir = None
-    if cfg:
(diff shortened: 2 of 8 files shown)


awcrosby (approved):

Spredzy:
This works as expected, but is not enough to fix  https://issues.redhat.com/browse/AAH-91.

In order to fix it, in addition to be able to specify `TMP_ROOT_DIR`  we need to be able to specify `ANSIBLE_LOCAL_TEMP`[1]

This is used by `ansible-doc` to do its duties. So, in a SELinux environment, we may want to specify an `ANSIBLE_LOCAL_TEMP` that has the proper labelling (vs. default `${HOME}/.ansible/tmp` which might not have it)

```
[root@ip-10-0-10-175 ec2-user]# strace ansible-doc file 2>&1 | grep tmp
stat("/root/.ansible/tmp", {st_mode=S_IFDIR|0700, st_size=6, ...}) = 0
mkdir("/root/.ansible/tmp/ansible-local-22084mwnRCW", 0700) = 0
```

Now there are two ways we can do this:

  * **The opinionated way**: Given we are already specifying `TMP_ROOT_DIR` we know that the `collections_path` will have proper labelling, so we can be opinionated and reuse that here. No new configuration variable to add.

```path
diff --git galaxy_importer/loaders.py galaxy_importer/loaders.py
index 6900703..a911cc1 100644
--- galaxy_importer/loaders.py
+++ galaxy_importer/loaders.py
@@ -229,6 +229,7 @@ class DocStringLoader():
         collections_path = '/'.join(self.path.split('/')[:-3])
         cmd = [
             '/usr/bin/env', f'ANSIBLE_COLLECTIONS_PATHS={collections_path}',
+            f'ANSIBLE_LOCAL_TEMP={collections_path}
             'ansible-doc',
             '--type', plugin_type,
             '--json',
```

  * **The more flexible way**: Just like `TMP_ROOT_DIR` add a new `ANSIBLE_LOCAL_TEMP` in `[galaxy-importer]` section

[1] https://docs.ansible.com/ansible/latest/reference_appendices/config.html#default-local-tmp

awcrosby:
Either approach sounds ok to me

galaxy_importer/collection.py
```diff
@@ -58,7 +58,10 @@ def import_collection(file, filename=None, logger=None, cfg=None):
 
 
 def _import_collection(file, filename, logger, cfg):
-    with tempfile.TemporaryDirectory() as tmp_dir:
+    tmp_root_dir = None
+    if cfg:
+        tmp_root_dir = cfg.tmp_root_dir
```
awcrosby, line 63: looking again at this, since we have a default config, `if cfg` will always resolve to true, recommend:
`tmp_root_dir = cfg.tmp_root_dir`

galaxy_importer/loaders.py
```diff
@@ -226,9 +227,13 @@ def _get_plugins(self, plugin_dir):
         return plugins
 
     def _run_ansible_doc(self, plugin_type, plugins):
+        ansible_local_tmp = '~/.ansible/tmp'
+        if self.cfg:
+            ansible_local_tmp = self.cfg.ansible_local_tmp
```
awcrosby, line 232: same here, with `self.cfg` always present, this probably should be
```ansible_local_tmp = self.cfg.ansible_local_tmp or '~/.ansible/tmp'```

File: README.md [hunk 3]
```diff
@@ -47,6 +47,7 @@ RUN_ANSIBLE_LINT = True
 RUN_ANSIBLE_TEST = False
 ANSIBLE_TEST_LOCAL_IMAGE = False
 INFRA_OSD = False
+TMP_ROOT_DIR = None
 ```
```
awcrosby, line 51: probably worth adding the 2nd config option here

awcrosby (changes requested):

awcrosby (approved):

PR #105, opened 2020-11-30
Move ci testing and pypi publish to github actions

awcrosby:
No-Issue


.github/workflows/main.yml
```diff
@@ -1,20 +1,84 @@
...
+    - name: Install source package and dependencies
+      run: |
+        python -m pip install -e .[dev] 
+
+    - name: Lint with flake8
+      run: |
+        # stop the build if there are Python syntax errors or undefined names
+        flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+        # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
+        flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
```
cutwater, line 48: Why is `--max-line-length` set to 127?

cutwater (changes requested):

.github/workflows/main.yml
awcrosby: I pulled this from github's python workflow template... it is now fixed to say just `flake8 .` which uses our existing config file

cutwater (approved):

PR #114, opened 2021-02-10
Return execution_environment in importer result

bmclaughlin:
Issue: AAH-7
https://issues.redhat.com/browse/AAH-7

### DIFF
--- a/CHANGES/7.feature
+++ b/CHANGES/7.feature
@@ -1,0 +1,1 @@
+Move execution_environment from docs_blob to top-level importer result
--- a/tests/unit/test_loader_execution_env.py
+++ b/tests/unit/test_loader_execution_env.py
@@ -1,0 +1,76 @@
+# (c) 2012-2019, Ansible by Red Hat
+#
+# This file is part of Ansible Galaxy
+#
+# Ansible Galaxy is free software: you can redistribute it and/or modify
+# it under the terms of the Apache License as published by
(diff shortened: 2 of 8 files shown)


tests/unit/test_loader_execution_env.py
```diff
@@ -0,0 +1,79 @@
...
+    assert loader.data is None
+
+
+def test_no_directory():
+    loader = loaders.ExecutionEnvironmentLoader(collection_path="/does/not/exist/", log=logging)
+    assert loader.data is None
+
+
+def test_execution_env_data():
+    dir = os.path.dirname(os.path.dirname(os.path.dirname(__file__)))
```
awcrosby, line 65: can you use `with tempfile.TemporaryDirectory() as dir:` here like https://github.com/ansible/galaxy-importer/blob/master/tests/unit/test_builder_local_image.py#L99? that way the dir is removed and don't need to explicitly remove files

awcrosby (changes requested):

CHANGES/7.feature
```diff
@@ -0,0 +1 @@
+Return execution_environment to importer result
```
awcrosby, line 1: Recommend this say something like "Move execution_environment from docs_blob to top-level importer result" to be clear to those looking at changelog

awcrosby (approved):
One minor comment, otherwise LGTM

PR #116, opened 2021-02-19
Remove execution environment implementation and tests

bmclaughlin:
Issue: AAH-7
https://issues.redhat.com/browse/AAH-7


CHANGES/7.misc
```diff
@@ -0,0 +1 @@
+Remove no longer needed execution environment logic and tests.
```
awcrosby, line 1: this file should probably be a `.removal`

awcrosby (changes requested):

CHANGES/7.misc
bmclaughlin: didn't realize that was an option, thanks!

awcrosby (approved):

PR #117, opened 2021-02-27
Allow to install ansible(|-base|-core) separately

felixfontein:
Right now, galaxy-importer has a dependency on [ansible](https://pypi.org/project/ansible/), but in fact seems to be happy with any of ansible, ansible-core and ansible-base. Since pip does not allow to specify alternatives, my suggestion would be to drop the direct dependency on the ansible package, and allow to install ansible from extras. This is what ansible-lint is also doing since 5.0.0.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -15,7 +15,6 @@ classifiers =
 packages = find:
 
 install_requires =
-    ansible
     ansible-lint>=5.0.0,<6.0
     attrs>=19.3.0,<21
     bleach>=3.3.0,<4
@@ -33,6 +32,15 @@ dev =
     pytest-cov>=2.8.1,<3
     pytest_mock>=2.0.0,<3
     towncrier
+
+ansible =
+    ansible
+
+ansible-core =
+    ansible-core
+
+ansible-base =
+    ansible-base
 
 [options.package_data]
 galaxy_importer =


setup.cfg
```diff
@@ -15,7 +15,6 @@ classifiers =
 packages = find:
 
 install_requires =
-    ansible
```
awcrosby, line 18: I think we can make this a one line change making this `ansible-base`, I tested this earlier.

I was going to wait until `ansible-core` was out of pre-release, but we can go ahead and set to `ansible-base` for now to reduce install times.

awcrosby (changes requested):

setup.cfg
felixfontein: The problem with requiring ansible-base now is that it will make every dependent project (for example https://github.com/ansible-network/releases/blob/master/requirements.txt) have to follow the requirements of ansible-importer very closely. Eventually switching to ansible-core should be no problem (once that's properly released), but I don't think requiring ansible-base now is a good idea.

awcrosby: Makes sense. Well if we don't want to just wait for `ansible-core` coming out of pre-release... this PR will need an issue in our issue tracker + changelog entry to pass CI.

Also, any code using `galaxy-importer` will need to provide it's own `ansible(|-base|-core)` until we eventually just have `ansible-core` as the requirement. So we would temporarily need to add `ansible-base` or `ansible` to `galaxy_ng` as a direct dependency... if we merge this I can handle that. The alternative is to wait on `ansible-core`

felixfontein: @awcrosby now that ansible-core 2.11 has been released, it would be great if the dependency could be switched to that :)

felixfontein: (Though it might make sense to wait for Ansible 4.0.0 as well, so that packages that depend both on galaxy-importer and ansible will not have problems with both ansible-base and ansible-core being installed...)

felixfontein: (looking at https://github.com/ansible-network/releases/blob/master/requirements.txt :) )

PR #119, opened 2021-04-01
Use file_url from caller for remote storage

awcrosby:
Use the file_url from caller to support additional
pulp backend storage platforms in addition to AWS
which is currently supported.

Issue: AAH-431


bmclaughlin (approved):
LGTM

PR #121, opened 2021-05-17
Validate FILES.json and the files defined in it.

alikins:
https://issues.redhat.com/browse/AAH-403

### DIFF
--- a/galaxy_importer/collection.py
+++ b/galaxy_importer/collection.py
@@ -27,13 +27,14 @@ import requests
 
 from galaxy_importer import config
 from galaxy_importer import exceptions as exc
-from galaxy_importer.finder import ContentFinder
+from galaxy_importer.finder import ContentFinder, FileWalker
 from galaxy_importer import loaders
 from galaxy_importer import schema
 from galaxy_importer.ansible_test import runners
 from galaxy_importer.utils import markup as markup_utils
+from galaxy_importer.utils import chksums
+from galaxy_importer.utils import string_utils
 from galaxy_importer import __version__
-
 
 default_logger = logging.getLogger(__name__)
 
@@ -126,6 +127,43 @@ def _extract_tar_shell(tarfile_path, extract_dir):
     subprocess.run(args, cwd=cwd, stderr=subprocess.PIPE, check=True)
 
 
+def check_artifact_file(path_prefix, artifact_file):
+    """Check existences of artifact_file on fs and check the chksum matches
+
+    Args:
+        path_prefix (str): Any file path prefix we need to add to file paths in the
+            CollectionArtifactFile artifact_file
+        artifact_file (CollectionArtifactFile): object with the expected info about
+            the file on the fs that will be checked.
+            This info includes name, type, path, and checksum.
+
+    Raises:
+        CollectionArtifactFileNotFound: If artifact_file is not found on the file system.
+        CollectionArtifactFileChecksumError: If the sha256sum of the on disk
+            artifact_file contents does not match artifact_file.chksum_sha256.
+
+    Returns:
+        bool: True if artifact_file check is ok, otherwise should raise exception
+    """
+    default_logger.debug('artifact_file: %s', artifact_file)
+
+    artifact_file_path = os.path.join(path_prefix, artifact_file.name)
+    if not os.path.exists(artifact_file_path):
+        msg = f"The file ({artifact_file.name}) was not found"
+        raise exc.CollectionArtifactFileNotFound(missing_file=artifact_file.name, msg=msg)
+
+    actual_chksum = chksums.sha256sum_from_path(artifact_file_path)
+
+    if actual_chksum != artifact_file.chksum_sha256:
+        err_msg = "".join([f"File {artifact_file.name} sha256sum should be ",
+                           f"{artifact_file.chksum_sha256} but the actual sha256sum ",
+                           f"was {actual_chksum}"])
+        default_logger.error(err_msg)
+        raise exc.CollectionArtifactFileChecksumError(err_msg)
+
+    return True
+
+
 class CollectionLoader(object):
     """Loads collection and content info."""
 
@@ -137,11 +175,28 @@ class CollectionLoader(object):
 
         self.content_objs = None
         self.metadata = None
+        self.file_manifest_file = None
         self.docs_blob = None
         self.contents = None
 
     def load(self):
-        self._load_collection_manifest()
+        # NOTE: If we knew the chksum for MANIFEST.json, we could check it here first
+        self.manifest = self._load_manifest()
+
+        self.metadata = self.manifest.collection_info
+
+        # The default name for 'file_manifest_file' is FILES.json
+        self.file_manifest_file = self.manifest.file_manifest_file
+
+        # load data from FILES.json
+        self.file_manifest = \
+            self._load_file_manifest(path_prefix=self.path,
+                                     file_manifest_file=self.file_manifest_file)
+
+        # check chksum for each file in FILES.json
+        # Note: Will raise exceptions on file_manifest / FILES.json errors
+        self._check_file_manifest(self.path, self.file_manifest, self.file_manifest_file.name)
+
         self._rename_extract_path()
         self._check_filename_matches_manifest()
         self._check_metadata_filepaths()
@@ -168,17 +223,107 @@ class CollectionLoader(object):
             requires_ansible=self.requires_ansible,
         )
 
-    def _load_collection_manifest(self):
+    def _load_manifest(self):
         manifest_file = os.path.join(self.path, 'MANIFEST.json')
         if not os.path.exists(manifest_file):
             raise exc.ManifestNotFound('No manifest found in collection')
+
+        default_logger.debug('manifest_file: %s', manifest_file)
 
         with open(manifest_file, 'r') as f:
             try:
                 data = schema.CollectionArtifactManifest.parse(f.read())
             except ValueError as e:
+                raise exc.ManifestValidationError(str(e)) from e
+
+            default_logger.debug('data: %s', data)
+            default_logger.debug('data.file_manifest_file: %s', data.file_manifest_file)
+            return data
+
+    def _load_file_manifest(self, path_prefix, file_manifest_file):
+        """Load CollectionArtifactFileManifest data from file_manifest_file
+
+        Args:
+            path_prefix (str): Any file path prefix we need to add to file paths in the
+                CollectionArtifactFile artifact_file
+            file_manifest_file (CollectionArtifactFile): object with info about the
+                FILES.json in the artifact. The info includes the name, type, path,
+                and chksum.
+
+        Raises:
+            ManifestValidationError: If there are any errors loading, parsing,
+                deserializing, or validating the data in file_manifest_file
+
+        Returns:
+            CollectionArtifactFileManifest: The data from file_manifest_file
+
+        """
+        default_logger.debug('file_manifest_file: %s', file_manifest_file)
+
+        check_artifact_file(path_prefix=path_prefix, artifact_file=file_manifest_file)
+
+        files_manifest_file = os.path.join(path_prefix, file_manifest_file.name)
+        default_logger.debug('files_manifest_file: %s', files_manifest_file)
+
+        with open(files_manifest_file, 'r') as f:
+            try:
+                file_manifest = schema.CollectionArtifactFileManifest.parse(f.read())
+            except ValueError as e:
                 raise exc.ManifestValidationError(str(e))
-            self.metadata = data.collection_info
+
+        return file_manifest
+
+    def _check_file_manifest(self, path_prefix, file_manifest, file_manifest_name):
+        """Check the file content described in file_manifest
+
+        Check the chksums for files.
+        Check for any missing files.
+        Check for any missing dirs.
+
+        Args:
+            path_prefix (str): Any file path prefix we need to add to file paths in the
+                CollectionArtifactFile artifact_file
+            file_manifest (CollectionArtifactFileManifest): Object with list of
+                CollectionArtifactFile items. The CollectionArtifactFile data
+                is used to check and validate each file.
+            file_manifest_name (str): The name of the file manifest (ie, "FILES.json")
+
+        Raises:
+            CollectionArtifactFileNotFound: If artifact_file is not found on the file system.
+            CollectionArtifactFileChecksumError: If the sha256sum of the on disk
+                artifact_file contents does not match artifact_file.chksum_sha256.
+
+        Returns:
+            bool: All the items in file_manifest were found and valid
+        """
+
+        for artifact_file in file_manifest.files:
+            if artifact_file.ftype != 'file':
+                continue
+
+            check_artifact_file(path_prefix=path_prefix,
+                                artifact_file=artifact_file)
+
+        # check the extract archive for any extra files.
+        filewalker = FileWalker(collection_path=path_prefix)
+        prefix = path_prefix + '/'
+        found_file_set = set([string_utils.removeprefix(fp, prefix) for fp in filewalker.walk()])
+
+        file_manifest_file_set = set([artifact_file.name for artifact_file in file_manifest.files])
+        # The artifact contains MANIFEST.json and FILES.JSON, but they aren't
+        # in file list in FILES.json so add them so we match expected.
+        file_manifest_file_set.add('MANIFEST.json')
+        file_manifest_file_set.add(file_manifest_name)
+
+        difference = sorted(list(found_file_set.difference(file_manifest_file_set)))
+
+        if difference:
+            err_msg = \
+                f'Files in the artifact but not the file manifest: {difference}'
+            raise exc.FileNotInFileManifestError(unexpected_files=difference,
+                                                 msg=err_msg)
+
+        return True
 
     def _rename_extract_path(self):
         old_ns_dir = os.path.dirname(self.path)
@@ -274,6 +419,7 @@ class CollectionLoader(object):
         )
 
     def _check_metadata_filepaths(self):
+        # NOTE: This may be redundant if _check_file_manifest() looks for missing files
         paths = []
         paths.append(os.path.join(self.path, self.metadata.readme))
         if self.metadata.license_file:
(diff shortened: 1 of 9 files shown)


galaxy_importer/collection.py
```diff
@@ -137,11 +138,13 @@ def __init__(self, path, filename, cfg=None, logger=None):
 
         self.content_objs = None
         self.metadata = None
+        self.file_manifest_file = None
         self.docs_blob = None
         self.contents = None
 
     def load(self):
         self._load_collection_manifest()
+        self._load_collection_file_manifest()
```
awcrosby, line 147: I was confused with `file_manifest_file` thinking it was `MANIFEST.json` not `FILES.json`

galaxy_importer/collection.py
```diff
@@ -173,12 +176,56 @@ def _load_collection_manifest(self):
...
+        default_logger.debug('manifest_file: %s', manifest_file)
+
         with open(manifest_file, 'r') as f:
             try:
                 data = schema.CollectionArtifactManifest.parse(f.read())
             except ValueError as e:
                 raise exc.ManifestValidationError(str(e))
             self.metadata = data.collection_info
+            default_logger.debug('data.file_manifest_file: %s', data.file_manifest_file)
+            self.file_manifest_file = data.file_manifest_file
```
awcrosby, line 188: can we separate the analysis of `FILES.json` and `MANIFEST.json`?

galaxy_importer/schema.py
```diff
@@ -305,22 +305,72 @@ def _check_license_or_license_file(self):
...
+
 @attr.s(frozen=True)
 class CollectionArtifactManifest(object):
     """Represents collection manifest metadata."""
 
     collection_info = attr.ib(type=CollectionInfo)
     format = attr.ib(default=1)
-    file_manifest_file = attr.ib(factory=dict)
+
+    file_manifest_file = attr.ib(default=None,
```
awcrosby, line 325: > can we separate the analysis of `FILES.json` and `MANIFEST.json`?

ditto

galaxy_importer/collection.py
```diff
@@ -173,12 +176,56 @@ def _load_collection_manifest(self):
...
+
+    def _load_collection_file_manifest(self):
+        default_logger.debug('self.file_manifest_file: %s', self.file_manifest_file)
+
+        files_manifest_file = os.path.join(self.path, self.file_manifest_file.name)
+        if not os.path.exists(files_manifest_file):
+            raise exc.ManifestNotFound(f'No file manifest file ({self.file_manifest.name}) found in collection')
```
awcrosby, line 195: can this be a generic `ImporterError` or something different? if a collection has a `MANIFEST.json` I dont think we should throw a `ManifestNotFound` error

galaxy_importer/collection.py
alikins: Agreed. manifest, manifest_file, file_manifest, file_manifest_file kind of blur together.

`files_manifest_file` slightly better.
`files_file` seems slightly worse.

Trying to avoid  `files_json` or `files_json_file` since it's not necessarily a  json file (possibly in the future), but maybe the most obvious 'family feud rules' name.

galaxy_importer/collection.py
alikins: Probably. I kind of grabbed a reference to data.file_manifest_file here because currently there isn't a shared reference to the data from CollectionArtifactManifest() except for the 'collection_info' subset.

Could just return 'data' here  (or set self.artifact_manifest_data = data)

galaxy_importer/schema.py
alikins: Not sure I follow. By analysis do you mean the attr validators?

galaxy_importer/collection.py
alikins: Or do you mean, validate  CollectionArtifactManifest().collection_info first, and then CollectionArtifactManager().file_manifest_file  ?

galaxy_importer/schema.py
awcrosby: the `CollectionArtifactManifest` class is all about `MANIFEST.json`, this edit now makes it reference the `FILES.json`

galaxy_importer/collection.py
awcrosby: i would rather refactor `files_json` later if the format did change. having a method (and var names) called `_load_collection_file_manifest()` that is related to `FILES.json` _not_ the manifest file `MANIFEST.json` is confusing now to me and will be likely be more confusing in the future and make it hard to maintain.

I believe `manifest` var/methods should only reference `MANIFEST.json`

awcrosby: never thought about 'family feud rules' haha, but yeah i think it applies here

alikins: This is kind of further confused by MANIFEST.json not actually being a 'manifest' by any normal definition.

and of course a manifest file that is a manifest of manifest files...  brane hurt.

I think I have the logic and implementation mostly sorted now, so i'll revisit the names now.

galaxy_importer/collection.py
alikins: Hmm, yeah.  The latest version moves the check for the actual file into collections.check_artifact_file()
but the error raised from there will be a:

`CollectionArtifactFileNotFound("The file (FILES.json) was listed in the file manifest but it was not found")`

Which is kind of worse for the particular case of FILES.json.  I'll see if I can simplify

alikins: I removed the somewhat extraneous details about being listed in the file manifest from the error.

galaxy_importer/schema.py
alikins: file_manifest_file was always the section of Manifest.json that describes FILES.json.
Prev it was just a dict datastructure, but now it is a CollectionArtifactFile instance.

galaxy_importer/collection.py
```diff
@@ -173,12 +232,105 @@ def _load_collection_manifest(self):
...
+
+        difference = sorted(list(found_file_set.difference(file_manifest_file_set)))
+        default_logger.debug('found_file_set.difference(file_manifest): %s',
+                             pf(difference))
+
+        if difference:
+            err_msg = \
+                f'Files in the artifact but not the file manifest: {difference}'
+            raise exc.FileNotInFileManifestError(unexpected_files=difference,
+                                                 msg=err_msg)
```
alikins, line 330: Kind of want to extract this section to at least it's own method. It's kind of a hodgepodge of stuff at the moment.

Maybe even it's own module, so collection.py wouldn't need the FileWalker import or the string_utils import. 

hiding away some of the path munging and set() presentation also may make it clearer what `_check_file_manifest` does.

galaxy_importer/collection.py
```diff
@@ -173,12 +232,105 @@ def _load_collection_manifest(self):
...
                 data = schema.CollectionArtifactManifest.parse(f.read())
             except ValueError as e:
                 raise exc.ManifestValidationError(str(e))
-            self.metadata = data.collection_info
+
+            default_logger.debug('data: %s', data)
+            default_logger.debug('data.file_manifest_file: %s', data.file_manifest_file)
+            return data
+
+    def _load_collection_file_manifest(self, path_prefix, file_manifest_file):
```
awcrosby, line 247: I don't think we should introduce names `file_manifest` `FileManifest` when `manifest_file` is already in use, and people equate "manifest" to `MANIFEST.json` not `FILES.json`. Can these be changed to a synonym like `file[s]_listing` `file[s]_record` `file[s]_register` `files_json`?

If a var/class does not refer somehow to `MANIFEST.json` then I don't think it should be named "manifest"

galaxy_importer/collection.py
```diff
@@ -173,12 +232,105 @@ def _load_collection_manifest(self):
...
+
+        file_manifest_file_set = set([artifact_file.name for artifact_file in file_manifest.files])
+        # The artifact contains MANIFEST.json and FILES.JSON, but they aren't
+        # in file list in FILES.json so add them so we match expected.
+        file_manifest_file_set.add('MANIFEST.json')
+        file_manifest_file_set.add(file_manifest_name)
+
+        difference = sorted(list(found_file_set.difference(file_manifest_file_set)))
+        default_logger.debug('found_file_set.difference(file_manifest): %s',
+                             pf(difference))
```
awcrosby, line 324: s/pf/pprint.pformat would avoid a 2 char lowercase global

awcrosby (changes requested):

galaxy_importer/collection.py
```diff
@@ -173,12 +232,105 @@ def _load_collection_manifest(self):
...
+
+        for artifact_file in file_manifest.files:
+            if artifact_file.ftype != 'file':
+                continue
+
+            check_artifact_file(path_prefix=path_prefix,
+                                artifact_file=artifact_file)
+
+        # check the extract archive for any extra files.
+        ftw = FileWalker(collection_path=path_prefix)
```
awcrosby, line 312: not sure if ftw is a common acronym, but `filewalker` would be a standard name as an instance of the class

galaxy_importer/collection.py
```diff
@@ -173,12 +232,105 @@ def _load_collection_manifest(self):
...
+            ManifestValidationError: If there are any errors loading, parsing,
+                deserializing, or validating the data in file_manifest_file
+
+        Returns:
+            CollectionArtifactFileManifest: The data from file_manifest_file
+
+        """
+        default_logger.debug('file_manifest_file: %s', file_manifest_file)
+
+        check_artifact_file(path_prefix=path_prefix, artifact_file=file_manifest_file)
```
awcrosby, line 266: any reason this is called twice?, once in `CollectionLoader.load` right before this class method is called

galaxy_importer/collection.py
```diff
@@ -137,11 +176,31 @@ def __init__(self, path, filename, cfg=None, logger=None):
...
+
+        # The default name for 'file_manifest_file' is FILES.json
+        self.file_manifest_file = self.collection_manifest.file_manifest_file
+
+        # check the FILES.json file's chksum
+        check_artifact_file(path_prefix=self.path, artifact_file=self.file_manifest_file)
+
+        # load data from FILES.json
+        self.file_manifest = \
+            self._load_collection_file_manifest(path_prefix=self.path,
```
awcrosby, line 197: this is referred to as `file_manifest` elsewhere, maybe s/_load_collection_file_manifest/_load_file_manifest

galaxy_importer/collection.py
```diff
@@ -126,6 +130,41 @@ def _extract_tar_shell(tarfile_path, extract_dir):
     subprocess.run(args, cwd=cwd, stderr=subprocess.PIPE, check=True)
 
 
+def check_artifact_file(path_prefix, artifact_file):
+    """Check existences of artifact_file on fs and check the chksum matches
+
+    Args:
+        path_prefix (str): Any file path prefix we need to add to file paths in the
+            CollectionArtifactFile artifact_file
+        artifact_file (CollectionArtifactFile): object with info (name, path) of the file
```
awcrosby, line 139: would it be more clear to say `file object inside the collection artifact with info (name, path) of the file` to deferentiate  CollectionArtifactFile from the collection artifact file tarball?

galaxy_importer/schema.py
```diff
@@ -305,22 +305,72 @@ def _check_license_or_license_file(self):
...
+            artifact_file = CollectionArtifactFile(name=file_item['name'],
+                                                   ftype=file_item['ftype'],
+                                                   chksum_type=file_item.get('chksum_type'),
+                                                   chksum_sha256=file_item.get('chksum_sha256'))
+            new_list.append(artifact_file)
+
+    return new_list
+
+
+# FIXME: where do we want to load / validate this.
```
awcrosby, line 359: I see 6 FIXME and TODOs... they may be all valid as placeholder notes, but I doubt they will be addressed anytime soon after merge, anything we want to address before merge?

galaxy_importer/collection.py
alikins: I think I'll probably just remove those debug logging lines. Though they could be handy for... debugging.

galaxy_importer/collection.py
alikins: Agreed, lowercase 'filewalker' is better. Will update.

ftw is semi-common (there is a ftw() in std libc).

(I think I was thinking it would collide with `filewalker` module import, but never added a 'filewalker' module and didn't update the name)

alikins: updated

galaxy_importer/collection.py
alikins: Agreed, one of those calls is redundant. 

I'll remove the call in CollectionImport.load() so it is only used in _load_collection_file_manifest().

Unless we want CollectionLoader.load() to call `check_artifact_file(...FILES.json...)` in a separate step from reading/deserializing/parsing/validating the ds in FILES.json. I don't think we need that now unless we start running into failures where it is hard to tell what step failed.

alikins: updated

galaxy_importer/schema.py
alikins: The FIXMEs here are mostly about where/when/how we want to catch any exceptions raised and provide useful error messages,

# FIXME: Decide where we want missing info for FILES.json to raise an exception 
in CollectionArtifactManifest.parse() 

I don't think there is a unit test for a MANIFEST.json that doesn't have a file_manifest_file item in it's top level dict. That would cause a key error there that would bubble up... to at least CollectionLoader._load_collection_artifact().

I'll see if I can add a unit for _load_collection_artifact() that will trigger that case.


# FIXME: where do we want to load / validate this.
We do that in _load_collection_file_manifest() so I'll remove the FIXME

alikins: Removed `# FIXME: where do we want to load / validate this.`

galaxy_importer/schema.py
```diff
@@ -305,22 +305,71 @@ def _check_license_or_license_file(self):
...
+                                 validator=attr.validators.optional(
+                                     attr.validators.instance_of(CollectionArtifactFile)))
 
     @classmethod
     def parse(cls, data):
         meta = json.loads(data)
         col_info = meta.pop('collection_info', None)
         meta['collection_info'] = CollectionInfo(**col_info)
+
+        # FIXME: Decide where we want missing info for FILES.json to raise an exception
```
alikins: @awcrosby Can I assume a MANIFEST.json without a 'file_manifest_file' item should raise an exception? 

Updated to raise a `ValueError() from e` in parse, and the _load_collection_manifest() raises a
ManifestValidationError()  with chained exceptions to the ValueError and the KeyError

Added a unit test

alikins:
Updated the extraneous TODOs.

galaxy_importer/collection.py
alikins: Yeah, I'm inconsistent here. 

one extreme

``` python
# since this is returning a CollectionArtifactFileManifest.
#  But it is kind of redundant in a CollectionLoader() class...
collection_artifact_file_manifest =  self._load_collection_artifact_file_manifest(...)
```

given context other way
``` python
# Being in CollectionLoader partially implies we are reading a CollectionArtifact artifact file (tarball)
file_manifest = self._load_file_manifest(...)
```

Though, then I am inconsistent with

``` python
self.collection_manifest = self._load_collection_manifest()
```

I think I'm incline to leave it as is at the moment.

galaxy_importer/collection.py
alikins: Well, 'file object' implies a python file object to me. ie, `file_object = open('/tmp/somefile', 'r')`

I might rename CollectionArtifactFile -> CollectionArtifactFileInfo to imply it is not a 'file' itself, but info about a file (name, chksum, etc). Or maybe even 'tarfile' object like and CollectionArtifactFileMemberInfo() ?

alikins: Trying to avoid something like CollectionArtifactManifestFileMemberInfo because the CollectionArtifactFile class is used in CollectionArtifactManifest and CollectionArtifactFilesManifest.

I could ditch CollectionArtifact* entirely, but will likely also be RoleArtifacts soon.

galaxy_importer/collection.py
alikins: Could ditch the 'collection' from all here...

``` python
self.manifest = self._load_manifest()
self.file_manifest = self._load_file_manifest()
```

alikins: I picked https://github.com/ansible/galaxy-importer/pull/121#discussion_r647636656
updated

galaxy_importer/collection.py
awcrosby: maybe the docstring could say `file inside the collection artifact with info (name, path) of the file` ... i just want it to differentiate somewhere that `artifact_file` is not the artifact tarball file

alikins: @awcrosby Updated to be

``` python
    def _load_file_manifest(self, path_prefix, file_manifest_file):
        """Load CollectionArtifactFileManifest data from file_manifest_file

        Args:
            path_prefix (str): Any file path prefix we need to add to file paths in the
                CollectionArtifactFile artifact_file
            file_manifest_file (CollectionArtifactFile): object with info about the
                FILES.json in the artifact. The info includes the name, type, path,
                and chksum.
      <...>
      """
````

File: galaxy_importer/collection.py (hunk 10 above)
Comment (awcrosby) [reply galaxy_importer/collection.py]: > Could ditch the 'collection' from all here...
> 
> ```python
> self.manifest = self._load_manifest()
> self.file_manifest = self._load_file_manifest()
> ```

that seems like it would work

File: galaxy_importer/collection.py (hunk 11 above)
Comment (awcrosby) [reply galaxy_importer/collection.py:136]: @alikins i dont see the update in the files changed

Comment (alikins) [reply galaxy_importer/collection.py:136]: well, for check_artifact_file(), the artifact_file kinda isn't in the artifact, in the sense it has been extracted to fs already.
But I'll clarify the doc string.

Comment (alikins) [reply galaxy_importer/collection.py]:
``` diff
-        artifact_file (CollectionArtifactFile): object with info (name, path) of the file
+        artifact_file (CollectionArtifactFile): object with the expected info about
+            the file on the fs that will be checked.
+            This info includes name, type, path, and checksum.
```

awcrosby (approved):
LGTM

PR #122, opened 2021-06-08
Make `requires_ansible` in meta/runtime.yml mandatory

awcrosby:
Issue: AAH-538

### DIFF
--- a/galaxy_importer/loaders.py
+++ b/galaxy_importer/loaders.py
@@ -196,7 +196,7 @@ class RuntimeFileLoader():
             return
         requires_ansible = self.data.get('requires_ansible')
         if not requires_ansible:
-            return
+            raise exc.RuntimeFileError("'requires_ansible' in meta/runtime.yml is mandatory")
         if len(requires_ansible) > MAX_LENGTH_REQUIRES_ANSIBLE:
             raise exc.RuntimeFileError(
                 f"'requires_ansible' must not be greater than "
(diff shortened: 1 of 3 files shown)


galaxy_importer/loaders.py
```diff
@@ -196,7 +196,7 @@ def get_requires_ansible(self):
             return
         requires_ansible = self.data.get('requires_ansible')
         if not requires_ansible:
-            return
+            raise exc.RuntimeFileError("'requires_ansible' in meta/runtime.yml is mandatory")
```
alikins, line 199: Seems reasonable to me.

A possible alternative that honestly doesn't add much... If there were other reasons self.data['requires_ansible'] could fail maybe. But there aren't really other ways so no gain.

ie, leave your code as is unless this seems more useful 

``` python

try:
   requires_ansible = self.data['requires_ansible']
except KeyError as e:
   raise exc.RuntimeFileError("'requires_ansible' in meta/runtime.yml is mandatory") from e
```

alikins (approved):
LGTM. Added a tiny comment just to throw it out there but it doesn't really matter.

sshnaidm:
It's very bad communicated change (actually completely not) which broke all collections uploads and wasted a lot of time on redundant unnecessary efforts by investigation and fixing. When the value of this is actually nothing, `requires_ansible` is completely redundant in all collections.
Breaking interfaces silently is definitely a very bad idea.

PR #124, opened 2021-06-24
Fix requires_ansible check when no meta/runtime.yml

awcrosby:
Fail import when there is no meta/runtime.yml

Fixes a missed case from #122 and AAH-538

No-Issue

### DIFF
--- a/galaxy_importer/loaders.py
+++ b/galaxy_importer/loaders.py
@@ -193,10 +193,16 @@ class RuntimeFileLoader():
 
     def get_requires_ansible(self):
         if not self.data:
-            return
+            raise exc.RuntimeFileError(
+                "'requires_ansible' in meta/runtime.yml is mandatory, "
+                "but no meta/runtime.yml found"
+            )
         requires_ansible = self.data.get('requires_ansible')
         if not requires_ansible:
-            raise exc.RuntimeFileError("'requires_ansible' in meta/runtime.yml is mandatory")
+            raise exc.RuntimeFileError(
+                "'requires_ansible' in meta/runtime.yml is mandatory, "
+                "but 'requires_ansible' not found"
+            )
         if len(requires_ansible) > MAX_LENGTH_REQUIRES_ANSIBLE:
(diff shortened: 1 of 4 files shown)


tests/unit/test_collection.py
```diff
@@ -77,7 +77,7 @@
   "name": "FILES.json",
   "ftype": "file",
   "chksum_type": "sha256",
-  "chksum_sha256": "dc90402feea54f479780e067cba748559cb01bff52e6724a15264b9a55e8f000",
+  "chksum_sha256": "0c2e9514f54aaeda0605d8c9a60e71cab3e577113b58ecc65b839d2d1b86c216",
```
alikins, line 80: _<general comment on existing code, not a requested change...>_

Updating the tests to calculate this on the fly might be useful in the future. I didn't do it in an attempt to stay focused on the feature. Maybe it's better as is for explicitness, I'm not sure. It is a tiny bit annoying to update though.

awcrosby: Haven't had to update this much so should be ok, and wasn't a bad update

galaxy_importer/loaders.py
```diff
@@ -193,7 +193,7 @@ def _load(self):
 
     def get_requires_ansible(self):
         if not self.data:
-            return
+            raise exc.RuntimeFileError("'requires_ansible' in meta/runtime.yml is mandatory")
```
alikins, line 196: Probably fine. Maybe tweak the error msg to indicate the contents of meta/runtime.yml was empty?

alikins (approved):
LGTM. Small suggestion for updating the errmsg for the empty (or False-y) contents of runtime.yml, but that is not a blocker.

galaxy_importer/loaders.py
awcrosby: updated

felixfontein:
@awcrosby @alikins neither `meta/runtime.yml` nor `requires_ansible` are mandatory for collections. Why do you require them?

PR #125, opened 2021-06-24
Update CI commit check to match galaxy_ng workflow

awcrosby:
Update CI commit check to match galaxy_ng workflow, and fail CI when commit check is out of sync


alikins:
@awcrosby should the CHANGES/#.bugfix file be empty?

PR #129, opened 2021-08-01
Remove ansible-core from dependencies

sshnaidm:
Remove ansible-core from dependencies to allow this script to
be used with older ansible versions - ansible, ansible-base, etc.
This dep is not used anyway.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -15,7 +15,6 @@ classifiers =
 packages = find:
 
 install_requires =
-    ansible-core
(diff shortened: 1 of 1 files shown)


awcrosby:
`galaxy-importer` does call ansible-test which is provided via `ansible-core`. the dependency `ansible-list` also relies on bits provided by `ansible-core`.

if there was a clever way to keep `ansible-core` as a default dependency, but allow the use of `ansible` or `ansible-base`, that could work

PR #150, opened 2021-11-18
Fix permission issue on container collections dir

awcrosby:
Running as openshift job yields: mkdir: cannot create directory
'/ansible_collections/placeholder_namespace: Permission denied

No-Issue


drodowic (approved):
lgtm

PR #151, opened 2021-11-18
Add wget to ansible-test container

awcrosby:
It was removed by mistake, but is needed to download collection archive.

No-Issue


drodowic (approved):
lgtm

PR #158, opened 2022-03-15
Add support for ansible-lint to v6

ssbarnea:
Issue: AAH-1429

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -17,7 +17,8 @@ packages = find:
 install_requires =
     ansible-core
     ansible-builder>=1.0.1,<2.0
-    ansible-lint>=5.0.8,<6.0
+    ansible-lint>=5.4.0,<6.0; python_version<'3.8'
+    ansible-lint>=6.0.0,<7.0; python_version>='3.8'
     attrs>=21.2.0,<22
     bleach>=3.3.0,<4
     bleach-allowlist>=1.0.3,<2
--- a/CHANGES/1429.misc
+++ b/CHANGES/1429.misc
@@ -1,0 +1,1 @@
+Allow galaxy-importer to pull ansible-lint 6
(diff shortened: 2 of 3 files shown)


setup.cfg
```diff
@@ -17,7 +17,7 @@ packages = find:
 install_requires =
     ansible-core
     ansible-builder>=1.0.1,<2.0
-    ansible-lint>=5.0.8,<6.0
+    ansible-lint>=6.0.0,<7.0
```
Spredzy, line 20: Why the jump to `>=6.0` and not leave `>=5.0` ? Any technical reason ?

ssbarnea: They diverged considerably and will produce different results. We can keep it like this but you should be aware that ansible-lint upstream will not provide support for 5.x.

Ideally, you should look into removing support for py37.

awcrosby:
@ssbarnea the ci `build-and-test` fail issue that was at HEAD has been resolved. If you rebase then ci should run as expected.

awcrosby:
@ssbarnea I clicked rebase on this PR and the tests are passing, For commit check failure take a look at https://github.com/ansible/galaxy-importer/blob/master/PROCESS.md#issues-and-pull-requests

setup.cfg
```diff
@@ -17,7 +17,8 @@ packages = find:
 install_requires =
     ansible-core
     ansible-builder>=1.0.1,<2.0
-    ansible-lint>=5.0.8,<6.0
+    ansible-lint>=5.4.0,<6.0; python_version<'3.8'
+    ansible-lint>=6.0.0,<7.0; python_version>='3.8'
```
awcrosby, line 21: @ssbarnea I think we should keep it so galaxy-importer supports one range of ansible-lint

awcrosby (changes requested):

awcrosby:
closing via #164

PR #160, opened 2022-03-22
Prevent fail-fast when running build-and-test

ssbarnea:
We only have 3 jobs running in the matrix and the default fail-fast
strategy makes it very hard to spot the failure cause, and impossible
to compare with others.


awcrosby:
This PR looks good, but it needs an issue link or `No-Issue` on its own line in the commit to pass CI (since this is small `No-Issue` is fine)

cidrblock:
I went to add the issue to keep this moving forward, but issues are not enabled for this project.

thedoubl3j:
@cidrblock jira issue, not GH issue. but since this is small I would just `git commit --amend` and then on one of the commits, just pass `No-Issue`

awcrosby:
@ssbarnea I don’t think adding a single line in a commit message is a deterrent. The commit simply needs `No-Issue` added in a single line, and all commits need to be signed.

PR #166, opened 2022-04-01
error if a changelog file is not present at the root of the directory

thedoubl3j:
log an error to the output if a changelog file is not present. feature signed off from community and will not fail an import. jira is [here](https://issues.redhat.com/browse/AAH-1460)

will squash commits before merge.


tests/unit/test_loader_collection.py
```diff
@@ -101,6 +101,13 @@
    "chksum_sha256": "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855",
    "format": 1
   },
+  {
+   "name": "CHANGELOG.rst",
```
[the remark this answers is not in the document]
thedoubl3j: awesome, is there anyway to find that out etc? saw some being spat out in the logs but wasn't sure if there was a way I could see it before as to prevent failures etc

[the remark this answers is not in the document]
thedoubl3j: answered my own question, please ignore

awcrosby (changes requested):
To pass the failing CI check you need `Issue: AAH-####` on a single line inside at least one of your commits, and all commits need to be signed.

tests/unit/test_loader_collection.py
```diff
@@ -182,18 +191,24 @@ def readme_artifact_file(request):
 
 @pytest.mark.sha256("e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855")
 def test_check_artifact_file(populated_collection_root, readme_artifact_file):
-    res = chksums_utils.check_artifact_file(populated_collection_root, readme_artifact_file)
+    res = chksums_utils.check_artifact_file(
+        populated_collection_root, readme_artifact_file
+    )
```
awcrosby, line 196: we want to keep this formatted as-is. `make lint/format/black` should revert these changes automatically.

galaxy_importer/loaders/collection.py
```diff
@@ -115,6 +116,15 @@ def _check_ansible_test_ignore_files(self):  # pragma: no cover
                 line_count = len(f.readlines())
             self.log.warning(IGNORE_WARNING.format(file=ignore_file, line_count=line_count))
 
+    def _check_collection_changelog(self):
+        """Log an error when a CHANGELOG file is not present in the root of the collection."""
+        CHANGELOG_ERROR = "CHANGELOG.rst file not found at top level of collection."
```
awcrosby, line 121: There is no tests to see if `CHANGELOG.rst file not found at top level of collection.` gets into the logs (the edits to `tests/unit/test_loader_collection.py` only test missing file logic). I'm not sure if any of the existing unit tests do that which can be used as an example, but we probably should have that here.

awcrosby (changes requested):

tests/unit/test_loader_collection.py
thedoubl3j: I can re run it but I ran it before I pushed my final commit @awcrosby so i thought it would have caught that

galaxy_importer/loaders/collection.py
thedoubl3j: I am not seeing anything that is exposing the log during the run, everything else is writing directly to the log in the test run then asserting its there. Is there an object or something that is storing the log that we can compare against? I am not seeing this done is the other expected areas.

awcrosby: `caplog` can be used for that https://github.com/ansible/galaxy-importer/blob/df274d4021de1314562400c3a4dd4eb657520a13/tests/unit/test_runner.py#L107

thedoubl3j: ah so my understanding of caplog was for use of capturing `pytest` things only. will poke around and see what we can produce.

tests/unit/test_loader_collection.py
thedoubl3j:
```make lint/format/black
reformatted tests/unit/test_loader_collection.py

All done! ✨ 🍰 ✨```

File: tests/unit/test_loader_collection.py [hunk 4]
```diff
@@ -429,6 +429,18 @@ def test_missing_readme(populated_collection_root):
     assert "README.md" == excinfo.value.missing_file
 
 
+@mock.patch("galaxy_importer.collection.CollectionLoader._build_docs_blob")
+def test_changelog_fail(_build_docs_blob, populated_collection_root, caplog):
+    _build_docs_blob.return_value = {}
+
+    data = CollectionLoader(
```
awcrosby, line 436:
```suggestion
    CollectionLoader(
```
I forgot to run `make lint` on my code changes, this should fix the one CI check. I think you'll need to rebase and sign to get other CI check to pass.

PR #168, opened 2022-05-06
Use file objects without fs persistance

mdellweg:


galaxy_importer/collection.py
```diff
@@ -129,26 +130,21 @@ def _import_collection(file, filename, file_url, logger, cfg):
...
-
-        _extract_archive(tarfile_path=filepath, extract_dir=extract_dir)
+        _extract_archive(fileobj=file, extract_dir=extract_dir)
 
         data = CollectionLoader(extract_dir, filename, cfg=cfg, logger=logger).load()
         logger.info("Collection loading complete")
 
         ansible_test_runner = runners.get_runner(cfg=cfg)
         if ansible_test_runner:
+            filepath = file.name
```
awcrosby, line 140: The `OpenshiftJobTestRunner` needs to download the file from inside its image, will this still be ok?:
https://github.com/ansible/galaxy-importer/blob/2c0557d87a5dc5199a47fe841b4d2bd41366615a/galaxy_importer/ansible_test/runners/openshift_job.py#L87-L89

awcrosby: `OpenshiftJobTestRunner` is only used for console.redhat.com. We need to continue supporting the console.redhat.com use case (which the storage is public s3, not private). Whatever we pass to `OpenshiftJobTestRunner` needs to include a url that we can then pass to the container to download:

https://github.com/ansible/galaxy-importer/blob/085ebb2d62e094f93b6298ab27d232d7345b40d9/galaxy_importer/ansible_test/container/entrypoint.sh#L3-L6

Maybe with this PR we no longer need `_get_pulp_archive_url`

mdellweg: OK, i tried to understand all the intricacies here once again, and maybe we are fine. It's just _very_ confusing when the `file` argument is supposed to be file object or a django-storage file-like object. But as we still pass the file object handed down to the runner (and just additionally provide a local copy in `filepath`, that workflow should continue to work.

awcrosby: Agreed it is confusing, I think we can improve this. It feels like we could pass the pulp archive download url into the `ansible_test_runner`. Then we could pass the file object as-is to `ansible_test_runner`.

```
            ansible_test_runner(
                dir=tmp_dir,
                metadata=data.metadata,
                fileobj=file,
                download_url=_get_pulp_download_url(),
                logger=logger,
            ).run()
```
If you can suggest the best way to get that download url in this module based on the fileobj, I can do the leg work of updating the 3 ansible_test_runner classes (maybe I could make a PR rebased on top of this and we can merge together after they are both complete?)

mdellweg: Part of the complication is that pulp cannot provide a url for an artifact in all configurations. It only serves published content, and this all is part of a check before even turning it into content.
But for the sake of this PR, can we analyse, whether it serves that usecase, and after merging enumerate all the possible configuration combinations and write unittests for them to pin down the constraints before further refactoring?

awcrosby: The only time we will run `OpenshiftJobTestRunner` is under the configuration where we expect that we can get a download url from pulp. So if we can keep that working within this PR then we can merge this. Afterwards +1 to ensuring tests and then potential refactoring.

mdellweg: As we are passing along the filelike object, which in case of the s3 installation provides `.storage.url` i'm convinced it will continue to work that way.

mdellweg:
@awcrosby What do you think about the addition commit (to be squashed later...) of passing the already provided url down to the runners framework?

awcrosby (approved):
LGTM

PR #170, opened 2022-05-09
'requires_ansible' does not need to be semver

alikins:
Import was checking if the requires_ansible field was
valid semver. But the format used by ansible for versioning
is not valid semver.

Instead it is based on pythons builting
packagaging.specifiers.Specifier and SpecifierSet

This changes replaces the validation of 'requires_ansible'
with one based on SpecifierSet.

Issue: AAH-981

### DIFF
--- a/CHANGES/981.bugfix
+++ b/CHANGES/981.bugfix
@@ -1,0 +1,1 @@
+Change 'requires_ansible' to use custom ansible ver spec instead of semver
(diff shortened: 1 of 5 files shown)


alikins:
Draft because it needs more tests. Potentially some runtime_file tests with specs that are semver but not packaging.specifier compliant.

Need to review and make sure that nothing actually compares the version to the requires_ansible spec, but I didn't see anything.

awcrosby (approved):

PR #186, opened 2022-10-31
Fixes to work with ansible-lint 6.8

awcrosby:
* Add project-dir param when calling ansible-lint
* Remove free-form syntax in test

No-Issue


jctanner (approved):
rubber stamp

PR #192, opened 2023-01-13
Add tox configuration

ssbarnea:
This change should make local testing much easier, especially with different versions of python. Tox also documents (`tox -l`) all possible commands with descriptions, something that the Makefile never supported.

No-Issue


awcrosby:
We are going to continue with our method of testing

PR #195, opened 2023-01-18
Remove ceiling from dependencies

ssbarnea:
- removes ceiling from dependencies, which basically are armed time-bombs
- adds a "lock" extra which includes the old ceilings, so this could
  be used by consumers more afraid of new libraries
- Fixes blocking issue affecting https://github.com/ansible/ansible-lint/issues/2874


ssbarnea:
Per discussions with maintainers I am dropping it.

PR #199, opened 2023-01-23
Add positive unit test for changelog

awcrosby:
No-Issue


awcrosby:
@alisonlhart I realized we never added a positive test for the changelog in the past, so this will complement the negative test we have.

alisonlhart:
LGTM :+1:

alisonlhart (approved):

PR #201, opened 2023-03-09
add support for parsing filter-type plugins via yaml files

philfry:
Hi,

I thought it'd be useful to fetch the filter plugin documentation as well.
Since filter files have no documentation blocks the documentation is stored in yaml files matching the filter name. This patch makes the importer read these yamls instead of python files when handling filter plugins.


philfry:
I guess this PR is no longer needed.

PR #202, opened 2023-03-30
Update ansible-lint upperbound to a specific version

awcrosby:
Issue: AAH-2214


awcrosby:
Note py3.8 fails with latest ansible-lint version, but https://github.com/ansible/galaxy-importer/pull/203 will move to using py3.9, py3.10, and py3.11 in ci.

jctanner (approved):

PR #209, opened 2023-04-04
Ansible lint collection level call and testing

alisonlhart:
Adding `ansible-lint` collection level call in `loaders/collection.py`. 

The full command is: `ansible-lint --profile production --exclude tests/integration/ --exclude tests/unit/ --parseable --nocolor`.

The `--parseable` flag mimics the output style of the rest of the log by eliminating additional new lines and building errors on one line. 

The `--exclude` flag for `tests/unit/` and `tests/integration/` prevent ansible-lint from flagging on `.yml` files we don't need to verify, as we do not check the structure or code quality of tests.

New config variables added: 
`run_ansible_lint_roles` - defaults to `False`
`run_ansible_lint_collection` - defaults to `True`

Adding tests for the collection level ansible-lint call for passing, erroring on a role, erroring on the `meta/runtime.yml` file, and checking the `stderr` logging style.  

AAH-2202


alisonlhart:
We could remove the `run_ansible_lint` config variable and just leave `run_ansible_line_collection` and `run_ansible_lint_roles`. Currently `run_ansible_lint` _and_ one of the other two has to be `True` for ansible-lint to run.

alisonlhart:
@awcrosby This is ready for review. I defer to you on the above comment, in case the current _run_ansible_lint_ config var should be removed.

README.md
```diff
@@ -63,6 +63,12 @@ ANSIBLE_LOCAL_TMP = '~/.ansible/tmp'
 
 - `LOCAL_IMAGE_DOCKER` - Set to `True` to run the `ansible-test` container image via Docker; otherwise, Podman will be used. Defaults to `False`.
 
+- `RUN_ANSIBLE_LINT_ROLES` - Set to `True` to run `ansible-lint` on roles (only). Defaults to `False`.
+
+- `RUN_ANSIBLE_LINT_COLLECTION` - Set to `True` to run `ansible-lint` on the entire collection. Defaults to `True`.
```
awcrosby, line 68: Calling ansible-lint for a role inside a collection was a workaround for when there was no ability to run ansible-lint at the collection level. Now we are implementing that feature. Thinking about this more, I’m leaning towards not maintaining that workaround. By linting the entire collection we are still linting the roles, we are just changing how they are called. People will see more lint output, but that does stop a collection from being imported, so we are not breaking compatibility. I think this is preferable to maintaining multiple ways to call and having more complex configuration options.

I propose we run ansible-lint at the collection level when `run_ansible_lint` is True, and remove running at the role-inside-a-collection level, and keep the config option as-is.

galaxy_importer/loaders/collection.py
```diff
@@ -99,6 +105,58 @@ def load(self):
...
+            cmd,
+            cwd=self.path,
+            encoding="utf-8",
+            stdout=PIPE,
+            stderr=PIPE,
+        )
+        proc.wait()
+
+        for line in proc.stdout:
+            if line.endswith("(warning)\n"):
```
awcrosby, line 149: what are the stdout possibilites that do not end in `(warning)`?

galaxy_importer/loaders/collection.py
```diff
@@ -99,6 +105,58 @@ def load(self):
...
+        ]
+        self.log.debug("CMD: " + " ".join(cmd))
+        proc = Popen(
+            cmd,
+            cwd=self.path,
+            encoding="utf-8",
+            stdout=PIPE,
+            stderr=PIPE,
+        )
+        proc.wait()
```
awcrosby, line 146: We'll need to replace `proc.wait()` with `proc.communicate()` to prevent deadlocks. Right now I believe deadlocks are very rare, but when calling lint on the entire collection instead of one role at a time, it will likely increase. You can replicate the change made in #213

galaxy_importer/loaders/collection.py
```diff
@@ -99,6 +105,58 @@ def load(self):
...
 
+    def _lint_collection(self):
+        """Log ansible-lint output.
+
+        ansible-lint stdout are linter violations, they are logged as warnings and errors,
+        depending on the rule level.
+
+        ansible-lint stderr includes info about vars, file discovery,
+        summary of linter violations, config suggestions, and raised errors.
+        Only raised errors are logged, they are logged as errors.
```
awcrosby, line 116: Originally there were a lot of extra lines in stderr that did not make sense to list in the galaxy-importer log, so we only logged those that started with `constants.ANSIBLE_LINT_ERROR_PREFIXES` - is this still the case?

README.md
alisonlhart: This is fine by Partner Engineering. So we'd fully remove the existing call in content.py?

galaxy_importer/loaders/collection.py
alisonlhart: `ansible-lint` has an internal tagging system for rules that determines which ones are a 'warning' level and which ones are 'error' level. The 'warning' level is usually reserved for rules that are 'experimental', aka new to the rule list. 
Here's an example of a 'warning' level lint output that would be a line in stdout:
```tests/sanity/ignore-2.9.txt:139: sanity[cannot-ignore]: Ignore file contains pylint:consider-using-f-string at line 139, which is not a permitted ignore. (warning)```

Here's an example of an `error` level lint violation in stdout:
```playbooks/test-playbook.yml:36: name[play]: All plays should be named.```

The `(warning)` string at the end of the line is the best way I found to determine what violations aren't full errors as that is consistent for all warnings.

File: galaxy_importer/loaders/collection.py (hunk 3 above)
Comment (alisonlhart) [reply galaxy_importer/loaders/collection.py]: Change replicated.

File: README.md (hunk 1 above)
Comment (alisonlhart) [reply README.md]: Refactored to remove the loader level call.

File: galaxy_importer/loaders/collection.py (hunk 2 above)
Comment (alisonlhart) [reply galaxy_importer/loaders/collection.py]: @awcrosby I agree! Though when I'm thinking about it now, I'm not sure we're ready to fail import fully on ansible-lint or ansible-test sanity errors yet, and having them labeled as "warning" might communicate to partners that "warning" related-issues do not need to be resolved, vs if they were labeled with "error". I'll default to your judgement on that - but I could see us getting feedback from partners about "Why do we need to fix this if its just a warning?"

File: galaxy_importer/config.py [hunk 5]
```diff
@@ -30,6 +30,7 @@
 class Config(object):
     """Configuration for galaxy-importer."""
 
+    # TODO: remove run_ansible_lint?
```
awcrosby, line 33: I think this todo can be removed?

alisonlhart: Good catch.

galaxy_importer/loaders/collection.py
awcrosby: We should probably route ansible-lint stdout into the log type of warning rather than error (that matches current functionality). As for those stdout that end in `(warning)`... those could either not be logged, or logged in the same way as lines without it.

galaxy_importer/loaders/collection.py
```diff
@@ -88,6 +92,8 @@ def load(self):
         self.contents = self._build_contents_blob()
         self.docs_blob = self._build_docs_blob()
         self.requires_ansible = loaders.RuntimeFileLoader(self.path).get_requires_ansible()
+        if self.cfg.run_ansible_lint:
+            self._lint_collection()
```
awcrosby, line 96: since this method logs text that is more open-ended and can have lots of output (and it is configurable), it might be best to call this at the end, after the more simple checks below.

alisonlhart: Agreed. :+1: I'll move it.

galaxy_importer/loaders/collection.py
alisonlhart: @awcrosby Now that you say that - just today, after we started talking about this, ansible-lint had a change that makes it highly unlikely that there will be any `(warning)` messages when calling the `--production profile`, so I think it's safe to remove that condition and set everything ansible-lint related to warning level.

awcrosby:
@alisonlhart this PR looks good, I also tested it in our galaxy dev stack. I think the only thing needed is to resolve the merge conflicts, then it can be merged.

galaxy_importer/loaders/collection.py
awcrosby: This looks ok as-is

alisonlhart:
Copying over from #217: 

PE's thought of how #209 (this), #217, and #218 's merge and release could go is:

1. Merge #217 and tag release 0.4.11 for pipeline release before AAP 2.4 date
2. After the 0.4.11 release, merge both #209 (this) and #218, and tag release 0.4.12 for pipeline release ~AAP2.4 release date.
The logic of this would be to get the EDA tests into the pipeline ahead of the 2.4 date, since we will be releasing EDA content onto Automation Hub before the 2.4 release date, and align the ansible-core 2.15 bump and ansible-lint pipeline update with the 2.4 release.

Or, we could do all three PRs in the ~AAP2.4 release, but it would make EDA testing more complicated in the beginning. 

@awcrosby

awcrosby:
@alisonlhart when you say EDA testing, do you mean running locally, or having CRC run them? We don't need a galaxy-importer release to have that in CRC.

I would like to get this PR merged, we have other work being done with linting legacy roles, and I want to minimize the merge conflicts

alisonlhart:
@awcrosby All these changes are intended for CRC only, no local testing enforcement (though these tests will run if you enable ansible_test_local_image, since it runs alongside ansible-test). 

We only need it in CRC. If no release is needed, then I'm totally fine with getting these few changes you requested in and then merging, then testing in stage asap.

alisonlhart:
Edited the above comment, as I was working on the EDA PR when I saw this and mixed them up. @awcrosby

awcrosby:
@alisonlhart sounds like lint can be merged now, since it won't get deployed to CRC via galaxy-importer until 13Jun or after.

The EDA tox work will update a container that can be updated sooner, after the summit freeze is over.

alisonlhart:
@awcrosby If all those changes in #217 can go in separately (is there a cherry-picked commit of galaxy-importer in the container build?), then it'd be fine with me to go ahead and merge this one.

awcrosby:
@alisonlhart any changes to https://github.com/ansible/galaxy-importer/tree/master/galaxy_importer/ansible_test are not used in a meaningful way via a `galaxy-importer` pypi release. On merge, those changes get built into an image on quay and can be added to CRC before or long after a pypi release is used in CRC.

For this PR, I think you just need to resolve the merge conflicts.

alisonlhart:
@awcrosby Fixed!

awcrosby (approved):

PR #210, opened 2023-04-05
Add import exception handling for ansible-builder shuffling.

jctanner:
No-Issue

### DIFF
--- a/galaxy_importer/loaders/collection.py
+++ b/galaxy_importer/loaders/collection.py
@@ -20,9 +20,9 @@ import os
 import re
 
 try:
-    import ansible_builder.introspect
+    from ansible_builder import introspect
 except ImportError:
-    import ansible_builder._target_scripts.introspect
+    from ansible_builder._target_scripts import introspect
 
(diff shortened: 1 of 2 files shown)


galaxy_importer/loaders/collection.py
```diff
@@ -19,7 +19,10 @@
 import os
 import re
 
-import ansible_builder.introspect
+try:
+    import ansible_builder.introspect
+except ImportError:
+    import ansible_builder._target_scripts.introspect
```
awcrosby, line 25: think would need to be refactored here and below to be `from ansible_builder._target_scripts import introspect` since below it uses

```
python_deps, system_deps = ansible_builder.introspect.process_collection(self.path)
```

awcrosby:
you'll need to change https://github.com/ansible/galaxy-importer/blob/master/setup.cfg#L50 from `95.57` to something like `95.50`, since more import lines impact coverage calculations

awcrosby (approved):

PR #213, opened 2023-04-10
Prevent deadlock in os process call

awcrosby:
Issue: AAH-2145


alisonlhart:
LGTM. I will need to make these changes in #209 as well.

alisonlhart (approved):

PR #215, opened 2023-04-20
Find and load eda extension plugins into content list

awcrosby:
Issue: AAH-2311


thedoubl3j (approved):
lgtm, like the exclusion as well if there are custom extensions as well.

galaxy_importer/constants.py
```diff
@@ -73,6 +83,8 @@ class ContentType(enum.Enum):
     TERMINAL_PLUGIN = "terminal"
     TEST_PLUGIN = "test"
     VARS_PLUGIN = "vars"
+    EDA_EVENT_SOURCE = EDA_EVENT_SOURCE_NAME
```
alisonlhart, line 86: Should EDA_EVENT_SOURCE = "event_source" and EDA_EVENT_FILTER = "event_filter" to indicate the type instead of being the full path?

awcrosby: In the future with custom extensions being used in core (and maybe being allowed by galaxy-importer) both `eda/plugins/event_sources` and `my_thing/plugins/event_sources` could exist. Also per core this value should be considered the plugin type, and that is what will be used with `ansible-doc -t`... `eda/plugins/event_sources` just like `module_utils` or `shell`. If the web UI display is desired to be different, galaxy can eventually utilize the `display_name` in the spec for `meta/extensions.yml`

alisonlhart: Got it!

alisonlhart:
According to @benthomasson and the EDA team, the EDA plugin structure will change to /extensions/eda/plugins/event_source (singular) and /extensions/eda/plugins/event_filter (singular). Ben can you confirm this here before this PR goes in?

galaxy_importer/constants.py
```diff
@@ -44,12 +44,22 @@
 MAX_TAGS_COUNT = 20
 NAME_REGEXP = re.compile(r"^(?!.*__)[a-z]+[0-9a-z_]*$")
 
+# For these extensions we support listing them in the galaxy contents list
+# In the future we may allow any extension to be listed, and call ansible-doc on it
+EDA_EVENT_SOURCE_NAME = "eda/plugins/event_sources"
+EDA_EVENT_FILTER_NAME = "eda/plugins/event_filters"
```
alisonlhart, line 50: These may need to be updated to:
EDA_EVENT_SOURCE_NAME = "eda/plugins/event_source"
EDA_EVENT_FILTER_NAME = "eda/plugins/event_filter"

alisonlhart: @benthomasson

benthomasson: It is singular

PR #219, opened 2023-05-22
Add support for legacy roles

EamonTracey:
This PR adds support for legacy roles, with 2 new CLI flags, `--legacy-role` and `--namespace`. I wrote a `LegacyRoleLoader` class and defined an output schema, `LegacyImportResult`, that represents what will be written to `importer_result.json`.

This PR is missing some key items, including proper tests and logging, but should be a good basis.


galaxy_importer/constants.py
```diff
@@ -53,6 +53,8 @@
     EDA_EVENT_FILTER_NAME,
 ]
 
+GITHUB_NAME_REGEXP = re.compile(r"^[a-z\d](?:[a-z\d]|-(?=[a-z\d])){0,38}$")
```
awcrosby, line 56: probably want to rename to say username

galaxy_importer/loaders/content.py
```diff
@@ -175,6 +175,22 @@ def _make_path_name(rel_path, name):
...
+        doc_strings=None,
+        cfg=None,
+        logger=None,
+        legacy=False,
+    ):
+        super().__init__(content_type, rel_path, root, doc_strings, cfg, logger)
+        self.legacy = legacy
+
+        if self.legacy:
+            self.path_name = self.name
```
awcrosby, line 192: another option... you could add a `**kwargs` to the parent class init signature along with `self.kwargs = kwargs`, then the child class can have additional named params, and you could then check for self.kwargs.get("legacy_role") or whatever in the child `__init__()` or child `load()`

either approach may be fine, but seeing the init params duplicated (and therefore needing to be maintained in 2 places) made me think of that

galaxy_importer/loaders/legacy_role.py
```diff
@@ -0,0 +1,125 @@
...
+        Only raised errors are logged, they are logged as errors.
+        """
+
+        self.log.info(f"Linting role {self.name} via ansible-lint...")
+
+        if not shutil.which("ansible-lint"):
+            self.log.warning("ansible-lint not found, skipping lint of role")
+            return
+
+        cmd = [
```
awcrosby, line 95: can you try adding `--profile production` and `--nocolor` to match https://github.com/ansible/galaxy-importer/blob/master/galaxy_importer/loaders/collection.py#L125 ... if they work when calling at the role level we can keep them consistent

galaxy_importer/loaders/legacy_role.py
```diff
@@ -0,0 +1,125 @@
...
+            self.log.error("Timeout on call to ansible-lint")
+            proc.kill()
+            outs, errs = proc.communicate()
+
+        for line in outs.splitlines():
+            self.log.warning(line.strip())
+
+        for line in errs.splitlines():
+            if line.startswith(constants.ANSIBLE_LINT_ERROR_PREFIXES):
+                self.log.error(line.strip())
```
awcrosby, line 125: can you add this at the end?
```
self.log.info("...ansible-lint run complete")
```

awcrosby (commented):
Looks really good, I added some comments, mostly code styling suggestions. I reviewed all but the schema validation and tests.

galaxy_importer/legacy_role.py
```diff
@@ -0,0 +1,33 @@
...
+    :param logger: Optional logger instance.
+    :param cfg: Optional config.
+
+    :raises exc.ImporterError: On errors that fail the import process.
+
+    :return: metadata of legacy role
+    """
+
+    if namespace is None:
+        raise exc.ImporterError("Importing legacy role requires explicit namespace")
```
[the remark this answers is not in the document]
EamonTracey: does it make sense to even have this check since `namespace` is a positional required arg?

galaxy_importer/schema.py
```diff
@@ -375,6 +376,138 @@ def parse(cls, data):
...
+                raise ValueError("metadata must be in the form of a yaml dictionary")
+            if "galaxy_info" not in metadata:
+                raise ValueError("galaxy_info field not found in metadata")
+            if not isinstance(metadata["galaxy_info"], dict):
+                raise ValueError("galaxy_info field must contain a dictionary")
+            try:
+                galaxy_info = LegacyGalaxyInfo(**metadata["galaxy_info"])
+            except TypeError as e:
+                raise ValueError("unknown field in galaxy_info") from e
+            dependencies = metadata.get("dependencies", list())
```
awcrosby, line 499: does this line effectively do this?
```
dependencies = metadata.get("dependencies")
dependencies = list(dependencies)
```

galaxy_importer/schema.py
```diff
@@ -375,6 +376,138 @@ def parse(cls, data):
...
+
+        if value is not None and len(value) > MAX_LENGTH_URL:
+            raise ValueError(f"url must not exceed {MAX_LENGTH_URL} characters")
+
+    @license.validator
+    def _validate_license(self, attribute, value):
+        """Ensure the license is not too long."""
+
+        if value is not None:
+            if len(value) > MAX_LENGTH_LICENSE * 2:
```
awcrosby, line 454: is the 2x multiplier here to match what was allowed in the `ansible/galaxy` codebase? might be better to make a new constant in this module

galaxy_importer/schema.py
```diff
@@ -375,6 +376,138 @@ def parse(cls, data):
...
+                raise ValueError(
+                    f"role license must not exceed {MAX_LENGTH_LICENSE * 2} characters"
+                )
+
+    @min_ansible_version.validator
+    @min_ansible_container_version.validator
+    def _validate_version(self, attribute, value):
+        """Ensure a version is not too long and abides by semantic versioning."""
```
awcrosby, line 462: this does not check if this is a valid semantic version, probably just need to correct the docstring

galaxy_importer/legacy_role.py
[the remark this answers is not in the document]
awcrosby: should be fine not having the check there

awcrosby (commented):
Added some comments

galaxy_importer/schema.py
EamonTracey: No, if there is no "dependencies" field in metadata, then `metadata.get("dependencies")` returns `None`. The subsequent call would then be `list(None)` which results in the runtime error `TypeError: 'NoneType' object is not iterable`

EamonTracey: The `list()` in my implementation replaces `None` as the default value for if there is no "dependencies" field in metadata, ensuring that dependencies will be a list

galaxy_importer/schema.py
EamonTracey: Yes, a new constant makes sense. I added this because MAX_LENGTH_LICENSE isn't long enough to handle `license (GPL-2.0-or-later, MIT, etc)` which is the default value after `ansible-galaxy role init` ... this isn't a valid SPDX license, but I decided not to validate for SPDX license (Collections do this) because plenty of legacy roles have non-SPDX licenses

galaxy_importer/main.py
```diff
@@ -98,6 +108,28 @@ def call_importer(args, cfg):  # pragma: no cover
...
     Method excluded from pytest unit test coverage, tests exist in tests/integration
     """
+    if args.legacy_role:
+        if args.file is None:
+            logger.error("Must supply the directory of the role")
+            return None
+        if args.namespace is None:
+            logger.error("Importing legacy role requires explicit namespace")
+            return None
+        namespace = args.namespace.lower()
```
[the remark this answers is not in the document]
EamonTracey: I will remove automatically lowercasing the namespace. In the loader, then, it makes sense to notify the user if a namespace would be valid if it lowercased, but still fail them

EamonTracey:
@awcrosby This is ready for another look

EamonTracey:
Updated to allow uppercase letters in namespaces. I used [this](https://github.com/shinnn/github-username-regex) for the github username regex, but it doesn't allow for uppercase letters, even though github allows them ... I believe we should either allow uppercase letters or automatically convert to lowercase

tests/unit/test_loader_legacy.py
```diff
@@ -0,0 +1,311 @@
...
+        "foo_bar-baz",
+        "3w6",
+        "$@#",
+        "this.role",
+        "docker!",
+        "big space",
+        "whyUpper",
+    ],
+)
+def test_load_name_regex(populated_role_root, invalid_name):
```
awcrosby, line 218: is there a reason you want to test this here, and then run the same tests in `test_schema_legacy`? ... i know this is testing the error is present when calling `LegacyRoleLoader()` vs the other test `LegacyGalaxyInfo()` but my first thought is to keep single DRY category of check closest to the source (in the schema tests)

awcrosby: ah, i see there is a regex match against the role name in both the loader and the schema... do we need that check in 2 places?

EamonTracey: The check in in 2 places because of how we decided to determine the role name. The regex match in the schema will validate the `role_name` metadata field if it exists. Then, in the loader, that becomes the role name. However, if there is no `role_name` metadata field, then the regex match in the schema doesn't run, and the role name inherits from the directory name. This then needs to be regex validated in the loader

EamonTracey: I can indent the regex match in `LegacyRoleLoader` underneath the `else` block, so that way, it runs only when necessary (`role_name` field not present) and the regex match never runs twice

EamonTracey:
@awcrosby To clarify, are we allowing uppercase namespaces in galaxy_ng?

galaxy_importer/schema.py
```diff
@@ -375,6 +381,153 @@ def parse(cls, data):
...
+            except TypeError as e:
+                raise ValueError("unknown field in galaxy_info") from e
+            dependencies = metadata.get("dependencies", list())
+        return cls(galaxy_info, dependencies)
+
+    @dependencies.validator
+    def _validate_dependencies(self, attribute, value):
+        for dependency in value:
+            if not isinstance(dependency, str):
+                raise ValueError("depdencies must be a list of strings")
```
awcrosby, line 526: spelling issue

tests/unit/test_schema_legacy.py
```diff
@@ -0,0 +1,195 @@
...
+        ["eamontracey.hello_role"],
+        ["geerlingguy.php", "eamontracey.hello_role"],
+    ],
+)
+def test_valid_dependencies(galaxy_info, valid_dependency):
+    metadata = LegacyMetadata(LegacyGalaxyInfo(**galaxy_info), valid_dependency)
+    assert metadata.dependencies == valid_dependency
+
+
+def test_invalid_dependency_type(galaxy_info):
```
awcrosby, line 172: Are there any other incorrect types you could add here under a `@pytest.mark.parametrize`? If you add string type like (where meta/main.yml has `dependencies: geerlingguy.java`) it doesn't error about type, but instead `ValueError: g must have namespace and name separated by '.'`

galaxy_importer/main.py
```diff
@@ -98,27 +107,43 @@ def call_importer(args, cfg):  # pragma: no cover
...
+            logger.error("Must supply the directory of the role")
+            return None
+        if args.namespace is None:
+            logger.error("Importing legacy role requires explicit namespace")
+            return None
         try:
-            data = collection.import_collection(fh, filename, logger=logger, cfg=cfg)
+            data = legacy_role.import_legacy_role(args.file, args.namespace, logger=logger, cfg=cfg)
         except ImporterError as e:
             logger.error(f"The import failed for the following reason: {str(e)}")
```
awcrosby, line 120: For collections, the call to schema catches the ValueError and makes it an ImporterError https://github.com/ansible/galaxy-importer/blob/56a7c1a37d5d144a93af94aca82c2df9b18b68e4/galaxy_importer/loaders/collection.py#L223-L226

This allows for expected errors to display cleanly with no traceback for the caller. one possible test you could write here for TDD is that an error like bad dependency should not end in a traceback.

expected:
```
(.env) awc:cisco_ios_test$ python -m galaxy_importer.main awcrosby-cisco_ios_test-1.0.4.0.tar.gz 
Importing with galaxy-importer 0.4.10
ERROR: The import failed for the following reason: Invalid collection metadata. Expecting 'version' to be in semantic version format, instead found '1.0.4.0'.
(.env) awc:cisco_ios_test$
```

legacy role exposes traceback:
```
(.env) awc:ansible-role-jenkins$ python -m galaxy_importer.main --legacy-role . --namespace awcrosby
Importing with galaxy-importer 0.4.10
ERROR: Unexpected error occurred: geerlingguy.java.wrong must have namespace and name separated by '.'
Traceback (most recent call last):
...
ValueError: geerlingguy.java.wrong must have namespace and name separated by '.'
(.env) awc:ansible-role-jenkins$
```

awcrosby:
> To clarify, are we allowing uppercase namespaces in galaxy_ng for legacy roles?

@EamonTracey That is correct

galaxy_importer/main.py
```diff
@@ -98,27 +107,43 @@ def call_importer(args, cfg):  # pragma: no cover
...
-        )
-
-    match = FILENAME_REGEXP.match(os.path.basename(args.file))
-    namespace, name, version = match.groups()
-    filename = collection.CollectionFilename(namespace, name, version)
-
-    with open(args.file, "rb") as fh:
+    if args.legacy_role:
+        if args.file is None:
+            logger.error("Must supply the directory of the role")
```
awcrosby, line 112: these two commands determin the role name as expected
```
(.env) awc:~$ python -m galaxy_importer.main --legacy-role code/ansible-role-jenkins --namespace awcrosby
(.env) awc:code$ python -m galaxy_importer.main --legacy-role ansible-role-jenkins --namespace awcrosby
```
but this does not
```
(.env) awc:ansible-role-jenkins$ python -m galaxy_importer.main --legacy-role . --namespace awcrosby
Importing with galaxy-importer 0.4.10
Determined role name to be .
```

galaxy_importer/main.py
```diff
@@ -98,27 +107,43 @@ def call_importer(args, cfg):  # pragma: no cover
...
-            logger=logger,
-            cfg=cfg,
-        )
-
-    match = FILENAME_REGEXP.match(os.path.basename(args.file))
-    namespace, name, version = match.groups()
-    filename = collection.CollectionFilename(namespace, name, version)
-
-    with open(args.file, "rb") as fh:
+    if args.legacy_role:
```
awcrosby, line 110: depending on the legacy-role parameter, ansible-lint sometimes gives no output

```
(.env) awc:~$ python -m galaxy_importer.main --legacy-role code/ansible-role-jenkins --namespace awcrosby
Importing with galaxy-importer 0.4.10
Determined role name to be ansible-role-jenkins
Linting role ansible-role-jenkins via ansible-lint...
...ansible-lint run complete
Legacy role loading complete
Importer processing completed successfully
(.env) awc:~$ cd code
(.env) awc:code$ python -m galaxy_importer.main --legacy-role ansible-role-jenkins --namespace awcrosby
Importing with galaxy-importer 0.4.10
Determined role name to be ansible-role-jenkins
Linting role ansible-role-jenkins via ansible-lint...
WARNING: ansible-role-jenkins/molecule/default/java-8.yml:3:3: syntax-check[specific]: 'apt_repository' is not a valid attribute for a Play
...
```

tests/unit/test_loader_legacy.py
awcrosby: do you see any issues with moving that check out of `LegacyRoleLoader` and into the `LegacyImportResult`?

That way even though we have 2 checks, each is done in the corresponding schema section where the data orginates: `LegacyImportResult` and `LegacyGalaxyInfo`

galaxy_importer/main.py
awcrosby: we don't need to support running this from any directory depth, especially since this is a legacy feature. my recommendation:

1. in the readme and in the `import_legacy_role` docstring, say you need to run this inside the parent directory of the role
2. have the importer fail if the directory depth is not equal to 1... that way users are not missing an error about their role name (since the importer thought it was `.`) or they are not missing lint output

tests/unit/test_loader_legacy.py
EamonTracey: That makes sense. Then, for the namespace regex match, that _could_  be moved to `LegacyImportResult` as well ... but I think by this metric, it still belongs in `LegacyRoleLoader` because namespace is passed directly to `LegacyRoleLoader`

galaxy_importer/main.py
EamonTracey: will be resolved by strictly allowing directory depth of 1

EamonTracey:
@awcrosby Changes have been made. Traceback from value errors raised during metadata parsing are suppressed and instead cleanly show the appropriate error message (as with collections). Also, you can run the importer on legacy roles only from the parent directory of the role. This fixes ansible-lint not running properly.

EamonTracey:
@awcrosby Also a separate issue: when a user runs `python -m galaxy_importer.main` without any arguments, an error with traceback is raised `TypeError: expected str, bytes or os.PathLike object, not NoneType`. Is this desired behavior? An alternative would be to display the help options if no arguments are passed

galaxy_importer/loaders/legacy_role.py
```diff
@@ -46,13 +46,18 @@ def load(self):
...
+        try:
+            result = schema.LegacyImportResult(
+                namespace=self.namespace,
+                name=self.name,
+                metadata=self.metadata,
+                readme_file=self.readme.name,
+                readme_html=self.readme_html,
+            )
+        except ValueError as e:
+            raise exc.ImporterError(str(e)) from e
```
awcrosby, line 58: The legacyrole schema is following the collection schema pattern for calling exceptions, which is good, but i do think there is a minor improvement we can make… to make metadata schema errors be raised by the actual module doing schema validation:

- Rename the `LegacyMetadataParserError` to `LegacyRoleSchemaError`
- In schema.py under classes `LegacyImportResult`, `LegacyMetadata`, `LegacyGalaxyInfo`, for every `raise ValueError` change to `raise LegacyRoleSchemaError`
- This allows us to remove the 2 try/except/raise `LegacyMetadataParserError`/`ImporterError` in legacy_role.py

@EamonTracey what do you think about that?

awcrosby:
@EamonTracey about running without any arguments... good catch on that, but the main usage of this tool is not running from cli, and users only know how to run it by looking at the README which gives them the info they need on how to call, and if they hit an error it's obvious, so probably ok to keep as-is to keep complexity lower

awcrosby:
@EamonTracey there is one optional comment/suggestion, otherwise i think this LGTM. I recommend getting this to work fairly well with galaxy_ng in your galaxy_ng PR (which may already be the case) before we merge this. Whenever we feel it is fairly stable we can merge this PR and then fix any galaxy_ng integration bugs afterwards.

galaxy_importer/loaders/legacy_role.py
EamonTracey: I think that's reasonable ... why transform every `ValueError` into an `ImporterError` with `try/except/raise from` `instead of raising directly

awcrosby (approved):

PR #221, opened 2023-06-21
[WIP] Add support for .rst README and docs.

jctanner:
No-Issue

### DIFF
--- a/galaxy_importer/utils/markup.py
+++ b/galaxy_importer/utils/markup.py
@@ -17,9 +17,12 @@
 
 import collections
 import hashlib
+import io
 import mimetypes
 import os
 
+import docutils.core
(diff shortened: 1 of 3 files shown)


galaxy_importer/utils/markup.py
```diff
@@ -72,6 +71,8 @@ def get_html(doc_file):
     """
     if doc_file.mimetype == "text/markdown":
         return _render_from_markdown(doc_file)
+    elif doc_file.mimetype == "text/x-rst":
+        return _render_from_restructured_text(doc_file)
```
awcrosby, line 75: probably need to pull out the html sanitization from _render_from_markdown so it can be called on the unsafe html coming from the rst method

PR #224, opened 2023-06-27
Use permissive legacy role name regex

EamonTracey:
This regex matches all roles currently present on beta-galaxy.ansible.com which the old regex did not do.


jctanner (approved):
wfm

PR #225, opened 2023-06-29
[WIP] ansible-doc support filter and test plugins

EamonTracey:
Issue: AAH-1605

### DIFF
--- a/CHANGES/1605.feature
+++ b/CHANGES/1605.feature
@@ -1,0 +1,1 @@
+Support documentation of filter and test plugins inside a collection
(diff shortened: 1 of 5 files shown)


EamonTracey:
This seems to work out of the box. However, I tested against [community.general](https://galaxy.ansible.com/community/general) and found that ansible-doc failed on the filter plugin `community.general.hashids` because it [does not have documentation](https://github.com/ansible-collections/community.general/blob/main/plugins/filter/hashids.py). This causes ansible-doc to fail on ALL filter plugins. This might impact all filter types.

EamonTracey:
Ah, the issue is that the filter and test plugin documentation can be provided through adjacent YAML files.

jctanner:
replaced by https://github.com/ansible/galaxy-importer/pull/261

PR #231, opened 2023-07-13
Increase ansible-lint timeout

rjeffman:
No-Issue


rjeffman:
As ansible-lint adds more checks, it is timing out for bigger collections as [ansible-freeipa|https://github.com/freeipa/ansible-freeipa] when running galaxy-importer on Github runners.

PR #256, opened 2024-01-17
Increase ansible-lint to 6.22.1

traytorous:


alisonlhart:
@traytorous This is good so far! We'll need to do more testing to confirm if we need to adjust the lint command call with more ignores.

alisonlhart:
We'll also have to adjust some tests since the newer version of lint has updates where certain errors aren't being thrown anymore (this is intentional).

PR #257, opened 2024-01-18
Enable filter and test plugins doc parsing.

jctanner:
https://issues.redhat.com/browse/AAH-2854
https://issues.redhat.com/browse/AAH-2855

### DIFF
--- a/galaxy_importer/constants.py
+++ b/galaxy_importer/constants.py
@@ -24,12 +24,14 @@ ANSIBLE_DOC_SUPPORTED_TYPES = [
     "callback",
     "cliconf",
     "connection",
+    "filter",
     "httpapi",
     "inventory",
     "lookup",
     "shell",
     "module",
     "strategy",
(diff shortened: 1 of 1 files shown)


felixfontein:
This is producing import errors of collections, since it makes the wrong assumption that every `.py` file in `plugins/filters/` and `plugins/tests/` is a documentable test. Filter and test plugins work differently, you usually have a bunch of them defined in one Python file, and optionally you can provide documentation for them in YAML files.

See for example `time.py` in https://github.com/ansible-collections/community.general/tree/main/plugins/filter (most of the `to_xxx.yml` files belong to that Python file), or https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/filter/core.py, to which most of the YAML files in that directory belong.

The only good way to enumerate test and filter plugins is to use `ansible-doc --list --json --type filter/test foo.bar`.

PR #261, opened 2024-01-20
Use ansible-doc to list the plugins by type instead of crawling the directory

jctanner:
https://github.com/ansible/galaxy-importer/pull/257#issuecomment-1902077720

@felixfontein

### DIFF
--- a/galaxy_importer/loaders/doc_string.py
+++ b/galaxy_importer/loaders/doc_string.py
@@ -49,7 +49,9 @@ class DocStringLoader:
         for plugin_type in constants.ANSIBLE_DOC_SUPPORTED_TYPES:
             plugin_dir_name = constants.ANSIBLE_DOC_PLUGIN_MAP.get(plugin_type, plugin_type)
 
-            plugins = self._get_plugins(os.path.join(self.path, "plugins", plugin_dir_name))
+            # use ansible-doc to list all the plugins of this type
+            found_plugins = self._run_ansible_doc_list(plugin_type)
+            plugins = sorted(list(found_plugins.keys()))
 
             if not plugins:
                 continue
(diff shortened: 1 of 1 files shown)


galaxy_importer/loaders/doc_string.py
```diff
@@ -79,8 +79,35 @@ def _get_plugins(self, plugin_dir):
...
                 plugins.append(".".join(fq_name_parts))
+
         return plugins
 
+    def _run_ansible_doc_list(self, plugin_type):
+        """Use ansible-doc to get a list of plugins for the collection by type."""
+        collections_path = "/".join(self.path.split("/")[:-3])
+        cmd = [
+            "/usr/bin/env",
+            f"ANSIBLE_COLLECTIONS_PATHS={collections_path}",
```
felixfontein, line 90: `ANSIBLE_COLLECTIONS_PATHS` is deprecated and will be removed in ansible-core 2.19. You only need `ANSIBLE_COLLECTIONS_PATHS` for backwards compatibity with Ansible 2.9.
```suggestion
            f"ANSIBLE_COLLECTIONS_PATH={collections_path}",
```

felixfontein (commented):
First comment, I'll try to run this, but probably only tomorrow since I have other plans today :)

felixfontein:
This no longer produces an error, but it doesn't find the `to_xxx` filters from `time.py`, but instead finds:
```
            {
                "content_name": "time",
                "content_type": "filter",
                "doc_strings": null,
                "readme_file": null,
                "readme_html": null
            },
```
The code in `DocStringLoader.load` correctly finds the new plugins. The problem is that doc strings are only stored in plugins that are found, and `ContentFinder` again looks for files, which isn't really correct for test and filter plugins.

(While looking at the code: isn't
```python
        try:
            first = next(contents)
        except StopIteration:
            return []
        else:
            return itertools.chain([first], contents)
```
in https://github.com/ansible/galaxy-importer/blob/master/galaxy_importer/finder.py#L52-L57 simply equivalent to `return list(contents)`?)

jctanner:
Okay, to resolve the larger issue of the final output not including the plugins it's going to take a larger refactor of contentfinder and collectionloader. 

I think we should get this patch merged as is so we can can fix the zuul error.

drodowic (approved):

PR #264, opened 2024-02-02
Updating changelog check to include the docs/ dir

alisonlhart:
Issue: AAH-3083

Update the changelog check to look under the `docs/` dir as well as the collection root for a changelog file, in case maintainers want to display their changelog.md files in the Galaxy UI ("Documentation") tab, or just want to store their changelog files under `docs/`.

Also add the ability to turn this check off by adding a config value, defaulting to `True` since this check was already running on all imports previously. 

Also add a section to the README "Additional Notes", with info about .md files (or HTML render-able files) in the `docs/` dir displaying in the "Documentation" tab on Galaxy and Automation Hub.

PR #265, opened 2024-02-02
Expanding on config options in the README

alisonlhart:
Adding more config options to the README with an explanation of the option and the defaults for each value. 

Included the changelog config option from #264. 

No-Issue


alisonlhart:
Linting will pass after #264 is merged.

jctanner (approved):

PR #275, opened 2024-04-11
Updating ansible-core version to 2.16, python to 3.11, and base-test-container

chynasan:
Updating base-test-container in the Dockerfile to 5.10.0 per https://github.com/ansible/ansible/blob/stable-2.16/test/lib/ansible_test/_data/completion/docker.txt

Updating the version of ansible-core used in the Dockerfile to 2.16.0 to align with AAP 2.5 requirements.

Hold this until the AAP 2.5 release.

Issue: AAH-3156

### DIFF
--- a/CHANGES/3156.misc
+++ b/CHANGES/3156.misc
@@ -1,1 +1,1 @@
-Update ansible-test base image and ansible-core including Python bump from 3.9 to 3.11.
+Update ansible-test base image and ansible-core, including Python bump from 3.9 to 3.11.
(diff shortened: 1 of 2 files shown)


alisonlhart:
LGTM! Note for future reference: the Python 3.9 -> 3.10 increase is because ansible-core 2.16 requires Python 3.10 at minimum for the controller.

alisonlhart:
@jctanner There was some conversation about not locking to one specific version (aka 2.16.0). Do you think we should move forward with this bump for now and look into that more going forward?

jctanner (approved):

PR #329, opened 2025-03-03
Support plugins with sidecar documentation

sivel:
This PR adds support for plugins with sidecar documentation.  Sidecar documentation is found in a `<stem>.yml` file located adjacent to the plugin file.

There are effectively 2 enumeration code paths in importer:

1. ansible-doc enumeration
2. filesystem walking for `.py` and `.ps1` files

The `ansible-doc` enumeration was effectively thrown away, and only used for extracting docs where a `.py` or `.ps1` file was found.

This combines the lists from both for full coverage since `ansible-doc` doesn't yet support enumerating extensions (such as eda), and the filesystem walking doesn't enumerate sidecar. I didn't extend the file system walking, as it could pick up unintended files, and ultimately we should move to only relying on ansible-doc enumeration once it supports extensions.

The test utilizing `community.general` is a good reproducer, as it contains sidecar documentation for `filter` plugins.  Another good test collection is `microsoft.ad` as it uses a lot of sidecar docs.

### DIFF
--- a/tests/integration/test_eda.py
+++ b/tests/integration/test_eda.py
@@ -50,13 +50,13 @@ def test_eda_import(workdir, local_image_config):
 
     # the data should have all the relevant bits
     assert results["contents"] == [
-        {"content_type": "playbook", "description": None, "name": "hello.yml"},
-        {"content_type": "role", "description": "your role description", "name": "test_role"},
         {
             "content_type": "module",
             "description": "Upper cases a passed in string",
             "name": "upcase",
         },
+        {"content_type": "playbook", "description": None, "name": "hello.yml"},
+        {"content_type": "role", "description": "your role description", "name": "test_role"},
     ]
     assert results["docs_blob"]["contents"] != []
     assert results["docs_blob"]["collection_readme"]["name"] == "README.md"
--- a/galaxy_importer/constants.py
+++ b/galaxy_importer/constants.py
@@ -75,7 +75,7 @@ LEGACY_ROLE_NAME_REGEXP = re.compile("^[a-zA-Z0-9_-]{1,55}$")
 LEGACY_NAMESPACE_REGEXP = re.compile("^([a-zA-Z0-9.]+[-_]?)+$")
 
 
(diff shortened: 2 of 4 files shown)


tests/integration/test_eda.py
```diff
@@ -50,13 +50,13 @@ def test_eda_import(workdir, local_image_config):
...
-        {"content_type": "playbook", "description": None, "name": "hello.yml"},
-        {"content_type": "role", "description": "your role description", "name": "test_role"},
         {
             "content_type": "module",
             "description": "Upper cases a passed in string",
             "name": "upcase",
         },
+        {"content_type": "playbook", "description": None, "name": "hello.yml"},
+        {"content_type": "role", "description": "your role description", "name": "test_role"},
     ]
```
rochacbruno, line 60: Can a test be added with the specific documentation type?

Can we generate or mock a collection with eda extension stuff and assert on it?

IIUC this test change is just reordering right?

rochacbruno: nm, I saw #330 now

rochacbruno (approved):

tests/integration/test_eda.py
sivel: This isn't covering EDA, but this is just reordering.

The test changes in `tests/integration/test_collections.py` have been updated to make sure we are picking up the documentation for filter plugins that were missed previously.

In my other PR, instead of re-ordering, I've made the test perform a sort for comparisons:

https://github.com/ansible/galaxy-importer/pull/330/files#diff-bd15eaea92516e3fc24670f037e19ce66e9cf60187603a827e123bf43f5d49a7L52

The reason this was required was due to a change I made at https://github.com/ansible/galaxy-importer/pull/329/files#diff-9d660c36abac0799a93b9e07fee7edbd736f0121100c5c9dfb00ce7d7a699623R387 to sort the files to ensure a consistent ordering.

sivel: Cool, since both PRs touched this test, I've rebased #330.

PR #330, opened 2025-03-05
Add support for extracting docs from EDA plugins

sivel:
This PR aims to add the capability to `galaxy-importer` to extract docs from collection extensions, specifically in this case EDA plugins.

As of now, `galaxy-importer` utilizes `ansible-doc` to enumerate plugins and extract their documentation.  As "collection extensions" are foreign to `ansible-core`, the `ansible-doc` utility does not currently contain the functionality to perform this duty.

In 50c83a132ccb56115e95e5c75fad03accac33533 functionality was added to `galaxy-importer` to enumerate the extenion plugins, but had no mechanism to load the documentation.  A documentation format used for all other ansible plugins has since been adopted for EDA, and `ansible-doc` can be "abused" by supplying a directory via `-M` to indicate another module directory to consider as modules, which enables extracting the docs. This has to be done on a plugin by plugin basis, as it also contains all other actual modules that are part of ansible-core.

The long term goal as implied above is for `ansible-doc` to support collection extensions, and once it does the `ExtensionLoader._get_plugin_doc_strings` can be removed. As of now it has been implemented to call `ansible-doc` for each plugin that is found, and return the docs.  The `PluginLoader._get_plugin_doc_strings` implementation should automatically support the functionality needed once `ansible-doc` can enumerate the extensions.

To test this change, the `ansible.eda` collection can be used. This is the same collection as used in `test/integration/test_eda.py`.  This PR also updates the collection to 2.6.0 which has adopted the proper documentation formats.

### DIFF
--- a/galaxy_importer/collection.py
+++ b/galaxy_importer/collection.py
@@ -164,6 +164,13 @@ def _import_collection(file, filename, file_url, logger, cfg):
 
 def _extract_archive(fileobj, extract_dir):
     fileobj.seek(0)
+    _extract_kwargs = {}
+    if hasattr(tarfile, "data_filter"):
+        # Python added support for tarfile extraction filtering (PEP706) in
+        # response to CVE-2007-4559. This was introduced in Py3.12, although
+        # backported to other some earlier versions, and the default behavior
+        # will change in Python 3.14
+        _extract_kwargs["filter"] = "data"
     with tarfile.open(fileobj=fileobj, mode="r") as tf:
         for item in tf.getmembers():
             if item.name.startswith("/") or "../" in item.name:
@@ -175,4 +182,4 @@ def _extract_archive(fileobj, extract_dir):
                 )
                 if not link_target.startswith(os.path.abspath(extract_dir)):
                     raise exc.ImporterError("Invalid link target detected.")
-        tf.extractall(extract_dir)
+        tf.extractall(extract_dir, **_extract_kwargs)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -99,5 +99,5 @@ relative_files = true
 branch = true
 
 [tool.coverage.report]
-fail_under = 95.50
+fail_under = 90.00
 precision = 2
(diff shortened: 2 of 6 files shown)


rochacbruno:
@sivel conflicting with the previous PR, can you rebase please.

sivel:
> @sivel conflicting with the previous PR, can you rebase please.

Done.

pyproject.toml
```diff
@@ -97,5 +99,5 @@ relative_files = true
 branch = true
 
 [tool.coverage.report]
-fail_under = 95.50
+fail_under = 90.00
```
sivel, line 102: Just as a comment here, due to the added code, that was covered by integration tests, the coverage drops to something like 94.7.  But with the changes above to enable coverage for integration tests, units and integration combined are above 96%.

I didn't rework the `Makefile` but technically units and integration could be run together since they are run within the same workflow anyway.  But I've not done that here.

rochacbruno (approved):

galaxy_importer/collection.py
```diff
@@ -175,4 +178,4 @@ def _extract_archive(fileobj, extract_dir):
                 )
                 if not link_target.startswith(os.path.abspath(extract_dir)):
                     raise exc.ImporterError("Invalid link target detected.")
-        tf.extractall(extract_dir)
+        tf.extractall(extract_dir, **kwargs)
```
sivel, line 181: Completely unrelated, but when I switched the `test_eda` tests to be able to contribute to coverage and added capsys, I noticed the warning from tarfile that python 3.14 would change the behavior, and figured I could just address this quickly now.

rochacbruno: So, in this case, can a comment be added explaining it?

or maybe another conditional made instead of the `hasattr` e.g:

```py
_kwargs = {}
if sys.version_info >= (3, 12):
    _kwargs["filter"] = "data"
tf.extractall(extract_dir, **_kwargs)
```

So it becomes clear the reason for the conditional arg.

sivel: Yeah, I'll add a comment.  The `filter` functionality has been backported within RHEL python, so a `hasattr` is the recommendation.

sivel: Comment added, and variable renamed.

rochacbruno (approved):

PR #367, opened 2025-08-04
removed unnecessary ansible-lint rules for testing certified content

traytorous:
The Partner Engineering team has decided to exclude these tests from being ran on certified content.


gundalow (approved):
Thanks for this.
As a reminder, individual collection's CI pipeines may run `ansible-lint` directly which still test for these items, which is OK.
This PR is about reducing the noise in Hub's importer logs

PR #423, opened 2026-04-23
Fix ReDoS vulnerabilities in name validation regexes

lucasc017:
## Summary

- Fix polynomial backtracking (ReDoS) in `CONTENT_NAME_REGEXP` and `NAME_REGEXP` by removing redundant `+` quantifiers on first character classes that overlap with the following group
- Fix exponential backtracking (ReDoS) in `LEGACY_NAMESPACE_REGEXP` by flattening nested quantifiers into a linear-time equivalent pattern
- All three fixes are semantically identical to the original regexes (verified exhaustively against hundreds of thousands of test strings)

Resolves [AAP-71476](https://redhat.atlassian.net/browse/AAP-71476)

## Test plan

- [x] All existing unit tests pass (140 passed, 3 xfailed)
- [x] `CONTENT_NAME_REGEXP`: exhaustively verified 41,371 strings — zero behavioral differences
- [x] `NAME_REGEXP`: exhaustively verified 41,371 strings — zero behavioral differences
- [x] `LEGACY_NAMESPACE_REGEXP`: exhaustively verified 597,871 strings — zero behavioral differences
- [x] Performance: 10,000-char adversarial inputs complete in <1ms (previously would hang)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/galaxy_importer/constants.py
+++ b/galaxy_importer/constants.py
@@ -37,14 +37,14 @@ ANSIBLE_DOC_SUPPORTED_TYPES = [
 ANSIBLE_DOC_PLUGIN_MAP = {"module": "modules"}
 ANSIBLE_DOC_KEYS = ["doc", "metadata", "examples", "return"]
 ANSIBLE_LINT_ERROR_PREFIXES = ("CRITICAL", "ERROR")
-CONTENT_NAME_REGEXP = re.compile(r"^(?!.*__)[a-z_]+[0-9a-z_]*$")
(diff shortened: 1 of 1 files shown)


dragid10 (approved):
Can confirm that the old and new regexes are mathematically equivalent!

Low Care Amount: They technically aren't functionally equivalent because of the change in capture groups. Thats perfectly fine in this context, but in the future, if someone were trying to use `.groups()`, that wouldn't behave the same way. Not saying this is something to be fixed, but just wanted to leave some context in case this ever needs to happen in the future

bmclaughlin (approved):

alisonlhart (approved):

